PR #4, opened 2016-12-01
Update meeting types to include two which are used in Philadelphia.

FlipperPA:
Update meeting types to include two which are used in Philadelphia. Distinguish between 'Speaker Discussion' and 'Speaker Only', and add 'Variety / Chair's Choice'.


FlipperPA:
Howdy @joshreisner and @mrlynn - what do you think of these additions and slight change to the meeting types?

FlipperPA:
After discussion, we agree that we should reduce the number of available meeting types rather than continuing to grow it. Keeping the dropdown of types to frequently used ones makes the most sense.

PR #10, opened 2017-02-15
Adding user-defined header to meeting list (archive) page

johnfreed:
This is a feature I was looking for, and turned out not to be too hard to implement. The user can create an arbitrary header that displays above the meeting list. In my case I add the `<h1>` tag that is at the top of all the other pages on my site, as well as a Key to define Closed, Open, Chips, etc. Others can create their own.

### DIFF
--- a/includes/admin_import.php
+++ b/includes/admin_import.php
@@ -2,7 +2,7 @@
 	
 //import CSV file and handle settings
 function tmsl_import_page() {
-	global $wpdb, $tsml_types, $tsml_programs, $tsml_program, $tsml_nonce, $tsml_days, $tsml_feedback_addresses, $tsml_notification_addresses, $tsml_distance_units;
+	global $wpdb, $tsml_types, $tsml_programs, $tsml_program, $tsml_nonce, $tsml_days, $tsml_feedback_addresses, $tsml_notification_addresses, $tsml_distance_units, $tsml_archive_header;
 
 	$error = false;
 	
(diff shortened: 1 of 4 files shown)


joshreisner:
I don't want to make it easy for sites to put a bunch of stuff above the
meeting list. I don't think that improves the experience for end-users.

johnfreed:
I can understand that. It was a good exercise for me to understand the codebase better.

If someone runs across this thread in looking for a similar feature, it is possible to implement it manually by following the instructions in the FAQ to modify a copy of  archive-meetings.php.

napoland:
Hi Josh, I was actually thinking about a similar feature.

In our case we would like to add some text BELOW the meeting list.
Our site will be listing meetings from the Polish region. We would like to add below a link to the international website. Can you re-evaluate this feature, but maybe enable the text below the list, not above?

joshreisner:
Hey @napoland, like @johnfreed mentioned, you can customize the archive-meetings.php page to add some info. You can also modify your header or footer to add a message that is only displayed on your meeting archive page via PHP. 

Overall this seems a little out-of-scope for the plugin. It feels CMS-y when WordPress itself is such a CMS in its own right. Basically I don't like the idea of including an HTML editor on the Settings page.

What if we enabled widgets above and below the meeting list? Would that solve the issue?

napoland:
Yes, I think enabling the widgets should solve it.

PR #19, opened 2017-10-17
update to select # of meetings to display

lovethecodelabs:
updated includes/ajax.php & includes/widget.php to work with a new assets/js/closestmeetings.min.js to allow users to select # of meetings to display (Limit 1 widget instance per website).


joshreisner:
Provisionally accepting changes to complete the PR, subject to some refactoring.

PR #20, opened 2017-12-28
Added support for showing featured image on a meeting page

faridoon:
This feature allows users to upload a featured image to a meeting, which is then displayed above the map on single meeting pages.


joshreisner:
Thanks! Looks like you'd have to add a featured image to each meeting, rather than one per location. Is it better to have one per meeting, or one per location do you think?

faridoon:
I think it makes sense either way.

Featured image per meeting offers the most flexibility since some users may
want to upload pictures other than just the location. Think of it like
custom thumbnails indicating a theme, meeting type, meeting name, etc..

Featured image per location requires less time of admins. But in this case,
they are restricted to location pictures only.

By the way, you have done an amazing job with this plugin. I'll be happy to
contribute whenever possible.

PR #22, opened 2018-01-11
Fix translation issues

codifryed:
Some things were correctly translating, but most things were not.
Especially on the admin panel.

The issue seemed to be loading the load_plugin_textdomain in the
init hook in the file init.php doesn't work correctly.
Moved it to the file variables.php in the plugins_loaded hook.
The translations now load at the correct time in the
initialization sequence and tests are successful.

### DIFF
--- a/includes/init.php
+++ b/includes/init.php
@@ -11,9 +11,6 @@ function tsml_init() {
 	//run any necessary upgrades
 	tsml_upgrades();
 	
-	//load internationalization
-	load_plugin_textdomain('12-step-meeting-list', false, '12-step-meeting-list/languages');
-	
 	//meeting list page
 	add_filter('archive_template', 'tsml_archive_template');
 	function tsml_archive_template($template) {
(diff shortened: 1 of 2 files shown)


joshreisner:
Thanks @codifryed, could you describe the problems you were having a little more? Was there a plugin involved, or were you using translation files?

joshreisner:
According to [this](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/), you were 100% right to put this in the `plugins_loaded` hook.

codifryed:
Hi @joshreisner, the problems I was having was with .po and .mo translation files (files also from translate.wordpress.org) all of course based on the current .pot template. The translations on the backend weren't working at all, and the front end was sparadic, in varying places, etc. I did some tests and played around until it was clear that the translations were simply not being loaded correctly. Research showed some confusion between using the init or plugins_loaded, but when I read that post you mentioned there it was the nail in the coffin. Moved it (since I thought having it in two places would be confusing) and everything translated perfectly, back and front end. Tested with varying .po & .mo files and without and everything just worked. :-)

codifryed:
Also I tested it in a clean dev environment. No extra plugins/themes etc.

joshreisner:
Great, I appreciate the catch. It's in the latest version, committed to WP today.

codifryed:
saweet! thanks!

PR #24, opened 2018-01-30
Update shortcodes.php

XGhozt00:
TH should close with TH, not TD


joshreisner:
Thanks @XGhozt00, appreciate it. This is added to the WP version now in 2.16.17.

PR #33, opened 2019-03-11
Adding ability to specify post_status when pulling meetings

cdtoews:
I need some way to have meetings in the plugin that don't get published to web, but I can still print. 
I can accomplish by calling your function like:
$meetings = tsml_get_meetings(array(
    'post_status'  => array("publish","private")
  ));

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -876,9 +876,6 @@ function tsml_get_meetings($arguments=array(), $from_cache=true) {
 			'numberposts'		=> -1,
 		));
 
-		write_log(__LINE__);
-		write_log('nubmer of $posts:' . sizeof($posts));
-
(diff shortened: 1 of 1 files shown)


joshreisner:
Ok, I'll work this in to the next version, sounds good.

joshreisner:
Sorry, you'll see this function has actually been majorly refactored recently to support caching. It'll have to be done in a different way

joshreisner:
Try using the current Github version now (it's ahead of the WP repository version), and you should be able to do:

```
$meetings = tsml_get_meetings(array(
  'post_status' => array('publish', 'private'),
), false);
```

The `false` at the end is required, it means don't load from cache. (Only published meetings are going in the cache.)

Issue #47, opened 2019-11-20
Improve success/failure feedback when entering an address or updating a meeting


tech2serve:
**Is your feature request related to a problem? Please describe.**
Yes.

A number of sites using the 12-step-meeting-list plugin experience difficulties with geocoding certain addresses in their local area. With one example, the site’s geocoding worked during the initial creation of a meeting, but subsequent changes to the meeting are not possible because the geocoding fails and the `Update` button is not active. 

Other users have reported issues with the `Update` button being grayed out, and they’re not sure why. Subsequent investigation determines they have an address that is not seen as valid by Google, but this isn’t clearly displayed to the user on the admin page. Ofttimes copying and pasting the address from the maps.google.com page resolves the issue even though the addresses were fundamentally the same. 

Even when an incorrect address is the source of the problem, it is difficult for inexperienced users to realize the address they’ve entered is the problem based on the lack of information provided by the plugin. Likewise, when the address is confirmed valid but something else is causing a geocoding problem, no debug information is provided to the user. 

The lack of feedback makes it difficult to debug problems locally and remotely.

**Describe the solution you'd like**

When the address is geocoded successfully, the form should show a green check mark next to the input field. When the geocoding is not successful, a red “x” should be displayed, and additional error information should be displayed to help the user determine the corrective course of action.

The debugging information should appear near the address field so users are able to quickly capture and report in a support request or bug report.

**Describe alternatives you've considered**

- Automatically generating bug reports when geocoding fails.
- Add logging that could then be sent to the developers
- Instead of a green check, just have the field become dark gray until a successful geocode is completed.

**Additional context**
This will help troubleshoot problems and improve trust for the users.

joshreisner:
I can probably do this. I would like to review this with specific cases before work is done though.

In my many years of experience with this, the most common reason for a greyed out button was a javascript error from another plugin interfering with geocoding.

I am aware of a recent case also that had to do with the billing, which is a separate issue and may need to be addressed first.

tech2serve:
The case that I cannot solve yet is Ogden, Utah, with one specific address. As I chatted in #slack, we deleted the meetings in question, then reentered them. That worked fine until we tried to edit the meeting information and the `Update` is greyed. Ogden has 166 meetings in 47 locations; I can't say with complete certainty, but this appears to be the only validated issue. 

[...]

PR #48, opened 2019-11-23
cleaning up geocoding logic

joshreisner:
this PR filters out any empty geocodes in the cache, and cleans up some faulty logic for #47 

in particular, there was one bug that would have caused an empty result if the API was over the query limit (and would have persisted that empty result)


joshreisner:
@tech2serve I think I was being superstitious and confirming a value that was in all likelihood already set this way. Just ruling things out basically.

Pretty sure that this won't have much of an effect though, and that the core issue is resolved by the logic refactor, especially when `OVER_QUERY_LIMIT` occurs

PR #50, opened 2020-01-07
adding a third PDF format (4x7) for meeting schedules

hmbrecords:
AA HMB Area 48 has contracted out the printing of meeting schedules that must be 4 in by 7 in in size and must be less than 80 pages in total.  This pull request fulfills these requirements as a third option for generating PDFs inside of the 12-step-meeting-list wordpress plugin.


joshreisner:
Hey @hmbrecords let me know if these changes work for you. It should accomplish the same thing, but eventually we'll want to decouple the font size from the page size so this is a small step toward that.

hmbrecords:
> Hey @hmbrecords let me know if these changes work for you. It should accomplish the same thing, but eventually we'll want to decouple the font size from the page size so this is a small step toward that.

@joshreisner thanks for tidying up the code in this PR, it looks fine to me.  I think in the future we definitely should allow users to set the size and font on their own somehow.

PR #52, opened 2020-01-12
fill meetings with next day if none left for current day

### LINKED ISSUE #51: Upcoming meetings widget fill meetings from next day
If no other meetings are upcoming the widget should fill meetings from next day. Currently it won't display any

pjaudiomv:
this will resolve #51, however it won't add to `?tsml-time=upcoming` endpoint. and it won't fill to total count set in the widget args. I'm not sure that is necessary.  would like to hear what other opinions are. because of the way the current_time wordpress function (function :)) using strtotime in conjunction with built in wp function seemed best option to get next day integer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .svn
 .sass-cache
 node_modules
+.idea
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -13,6 +13,12 @@ if (!function_exists('tsml_next_meetings')) {
         global $tsml_program, $tsml_programs;
         $arguments = shortcode_atts(array('count' => 5), $arguments, 'tsml_next_meetings');
         $meetings = tsml_get_meetings(array('day' => intval(current_time('w')), 'time' => 'upcoming'));
+        $next_day = date('w', strtotime('+1 day', current_time('timestamp')));
+        if (count($meetings) < $arguments['count']) {
+            $added_meetings = tsml_get_meetings(array('day' => $next_day));
+            $added_meetings = array_slice($added_meetings, 0, $arguments['count']);
+            $meetings = array_merge($meetings, $added_meetings);
+        }
         if (!count($meetings)) {
             return false;
         }
@@ -20,6 +26,7 @@ if (!function_exists('tsml_next_meetings')) {
         //usort($meetings, 'tsml_next_meetings_sort');
         $meetings = array_slice($meetings, 0, $arguments['count']);
         $rows = '';
+        $ret = '';
         foreach ($meetings as $meeting) {
             if (is_array($meeting['types'])) {
                 $flags = array();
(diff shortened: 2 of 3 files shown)


pjaudiomv:
I guess one of the questions is too, do we want to merge the arrays with meetings for both days or only display just tomorrows meetings if current days upcoming meetings is 0.

tech2serve:
Thanks for the work, and taking this issue on!

In my first test (using Theme 2020 and the sample data set we provide with the plugin), I'm noticing that even though there are upcoming meetings today, the widget shows those *plus* the first two meetings from the next day (Monday in this case). I think this is what you're referring to in the comment above?

I guess I'm not hard over on whether we only show today's or today + the next day, but we need to modify the view to make it clear.

pjaudiomv:
I agree, the view doesn't display the day so it def is not clear. so seems like we maybe we should add the day to the view.

joshreisner:
I have to question the premise here. I don't think it's desirable to show tomorrow's meetings if there aren't any today. I think the current behavior (only show meetings from today) is far clearer.

What if there aren't any more meetings today and there's only one tomorrow? Are you going to try to fill up the 'gap' by adding more meetings from the next day?

I don't think this code complexity is worth it.

tech2serve:
Hi @joshreisner, We've started a discussion on this in the slack channel if you're interested.

We've experienced a few users that have had difficultly (and claimed the widget was broken) because it wasn't displaying anything. @pjaudiomv is one of those people, so he took on trying to improve it. I believe others think this is worthwhile, but complexity is an issue.

Another option is to just display "No further meetings" and then click on "more" to get the complete list?

pjaudiomv:
I have no ticket to reference but the request originally came form Brian W on slack, and I tend to agree. I don't think upcoming meetings needs to be limited to current day, however I do agree that to keep filling the `gap` by continuously adding days is not ideal or worth it. As a consumer of the widget I would expect it to always display something for upcoming meetings whether that is today or tomorrow. Maybe making an option in widget to display tomorrow meetings if none are left for today (but if code complexity is a concern [which is def always a valid concern to consider] this would definitely add more to it) Having said that I also don't care either way if the consensus is to not pursue this issue.

FlipperPA:
I have to agree with @joshreisner here - explicit is better than implicit. If the interface says you're looking at Tuesday's meetings, then lets not confuse the issue and show Wednesday's meetings. If there are no more meetings for the day... there are no more meetings for the day. Click to tomorrow.

I've had to resist building features for squeaky wheels myself. It is tempting to give a loud voice more weight than the thousands of people who are not writing in. I think part of our job is to remember that for every squeaky wheel, there are likely many more users who would just be more confused by this change than showing them what the interface is saying we are showing them.

Perhaps we tweak the message to the end user to, "There are no more meetings today. Please select another day in the dropdown." Something a bit instructive on what to do?

pjaudiomv:
just a point of clarity here, we are talking specifically about the upcoming meetings widget and tsml_next_meetings shortcode, Which does not currently display what day meetings you are looking at, so that confusion should be nil. you aren't looking at a specific days meetings only that they are "upcoming".

pjaudiomv:
I would be happy to close this PR and open another to adjust message in the widget and link as needed if meeting count is 0 if that ends up being consensus.

tech2serve:
> I have to agree with @joshreisner here - explicit is better than implicit. If the interface says you're looking at Tuesday's meetings, then lets not confuse the issue and show Wednesday's meetings. If there are no more meetings for the day... there are no more meetings for the day. Click to tomorrow.
> 
> I've had to resist building features for squeaky wheels myself. It is tempting to give a loud voice more weight than the thousands of people who are not writing in. I think part of our job is to remember that for every squeaky wheel, there are likely many more users who would just be more confused by this change than showing them what the interface is saying we are showing them.
> 
> Perhaps we tweak the message to the end user to, "There are no more meetings today. Please select another day in the dropdown." Something a bit instructive on what to do?

Hi @FlipperPA, Thanks for chiming in. I think as @pjaudiomv has already clarified, the issue is with the upcoming meetings widget which isn't explicit about the day of the meetings listed. When no meetings are remaining, the widget just shows blank space, leaving some to wonder if it is working.

> I would be happy to close this PR and open another to adjust message in the widget and link as needed if meeting count is 0 if that ends up being consensus.
@pjaudiomv, I think that is the consensus.  I haven't seen anyone in Slack chime in, but I suspect most will be happy as long as we have the widget say something other than having a blank area. 

Again, thank you all for participating!

joshreisner:
@tech2serve @pjaudiomv Oh great! I'm glad this is something that's being discussed on Slack and not a drive-by kind of pr. I don't mean to stand in the way if there's a consensus being reached.

Some thoughts:
 
* we definitely want to keep the current behavior as an option; the widget was designed for the use-case of the long sidebar with multiple widgets (ala https://aasanjose.org/) in this case the fact that it doesn't output anything is the desired behavior, better than outputting "no upcoming meetings." but I can imagine that some themes might show a gap, for example in a three-column setup. i would vote for the default being the current behavior, but up to you.
* if there's a "no meetings" message it should be customizable, as the "Upcoming Meetings" header is. perhaps the ui could be "when there are no upcoming meetings, show a) nothing b) a custom message ____________?
* because it's on lots of pages, this has the potential to cause memory issues, so it's good to be careful here with the queries. it should be ok because we are caching meetings now, but it's something to keep an eye on
* the future might be to have a reactjs client-side thing here maybe with some way to store the meetings locally so they aren't transferred on every page request

pjaudiomv:
so it seems to be split 50/50 on the consensus here. I think what i'm going to do is close this PR, and open a new one with on option on the widget to display message for if there are no meetings left on current day. view more link would be changed accordingly. No current default behavior would be changed, new feature would be opt in. We can always revisit this another time if more people request this feature

Issue #56, opened 2020-01-19
restored meetings not displaying.


pjaudiomv:
**Describe the bug**
See [comment on #55](https://github.com/code4recovery/12-step-meeting-list/pull/55#issuecomment-576020308). after deleting a meeting and restoring it meeting does not automatically reappear. I have to hit update for it to show up again.

**To Reproduce**
1. Trash a meeting
2. verify meeting is gone on site.
3. Restore from trash
4. meeting is still gone.

**Expected behavior**
Meeting automatically reappears.

On version `3.4.22` Siteground/Bluehost shared hosting

pjaudiomv:
I think whats happening is once a meeting gets restored, it gets set to unpublished and to get it to published state again you just have to hit update. However there is no status displaying it as unpublished it still says published.

joshreisner:
For faster performance, we have a JSON cache of the meeting data. My guess is that when a meeting is un-trashed, that the cache is not getting updated. 

I think we probably just need to add a callback for the `wp_untrash_post` hook and run `tsml_cache_rebuild()` inside of it.

tech2serve:
I ran a couple of tests this morning and #57 seems to have solved this issue. @pjaudiomv are you good with closing this?

Next I'm going to figure out why my platform isn't showing the meeting list and then I'll start testing  #55 again. I may be a little slow this week as I'm involved in a NPS volunteer project.

pjaudiomv:
Yes sounds good. Closing as is resolved.

PR #57, opened 2020-01-20
making the trashed_post function work for both trashing and un-trashing

joshreisner:
This tweak should fix the behavior observed in #56 

Didn't see @pjaudiomv in the list of reviewers or I would have added them


pjaudiomv:
Looks good to me, on way to airport I'll test when I get there.

Issue #42, opened 2019-11-14
Automating JSON Feeds


tech2serve:
A [request was made](https://wordpress.org/support/topic/automating-json-feeds/) in the support forums to have the plugin automatically update JSON feeds coming into the parent/top level site.

wokat:
Hi all.

I've started work on this, as I did not find a matching branch:
https://github.com/aa-berlin/12-step-meeting-list/tree/add-cron-tasks

The state right now is:

- Little testing, but seems to work
- Naive refactoring that simply has the cron task do what the "Refresh"-Button does: Delete every meeting, reimport
- Cron task currently updates until it has run a certain hard-coded percentage of `max_execution_time`, but could probably just extend runtime?
- Added a locking mechanism; lock-breaking automatic based on start of execution and assumed max. possible run time of originator (assumption: has same predefined runtime as current process)
- Have no clue where to define "functions", so tried to leave unchanged code in place for sake of the history

I would suggest that we (or "I" :joy:) finish the following tasks before a pull-requests:

- Make schedule of import cron configurable
- Reuse local post-ids to allow for relating to posts, so update existing post of a meeting

tech2serve:
Hi @wokat! Great to see you here. @brianw-area24 in particular had some ideas on this and I'd suggest you two connect. Probably the best way is for you to join our [Slack channel](https://join.slack.com/t/code4recovery/shared_invite/zt-ci34rbcx-r8ROFpwIqL6AYYmGZFp5xw).  He is pretty good about responding in there. A couple of others may chime in as well.

wokat:
Current state (see PR above): 

* Cron jobs work! And faster than with full deletes/inserts before
* I would like to make the interval configurable still
* Updates are faster than before, but still slow; try adding some indexes and provide option to create them for the user
* No log viewing facilities yet (dump the html into a file like the meeting cache json?)
* Add a Break-Lock-Button? Import process is blocked based on start time and assumed max. duration derived from max_execution_time
* The Ajax-Version from before could poll the backend; now it just says "import running already" and stops without progress info
* Import is checking fields that seem to not be exported at the moment... Forgot which, though :sweat_smile:, or export might be stricter with contact fields than the meeting list display? Something like that...

Should we drop/postpone any of this for a first release?

wokat:
Hi @tech2serve,

I updated my PR: https://github.com/code4recovery/12-step-meeting-list/pull/66

[...]

PR #66, opened 2020-03-09
Automating JSON Feeds #42

wokat:
# JSON Feeds via Cron Job

See issue #42 

This branch implements regular, unsupervised imports of existing json data sources.

This changes the import behaviour from "delete all, reimport" to "mark as stale, update, delete untouched meetings".

Import batches now run until about 80% of script's max_execution_time is reached. **This can easily be changed to trying to extend execution time instead.** But author elected to not bother in this branch, as cheap hosting probably usually would thwart attempts at that. Lock breaking (see below) also depends on that for a cheap, transparent solution.

While an import batch is being processed, the import is locked. Automatic lock breaking occurs after script_max_execution_time, assuming that the predecessor of an import process either finished by then or failed.

Branch also adds a **"Parent Region"** that can be specified with each data source. Contained meetings' regions will be appended to that region, if not yet existing as a child of it or its ancestors.

This adds **two cron jobs**:

* tsml_cron_invalidate_data_sources 
  * Configurable schedule
  * Marks meetings as stale and sets the "import buffer" option
* tsml_cron_import_data_source_batch
  * Runs every 10 minutes, if cron spawning is being used
  * Runs every 1 minute, if cron spawning is switched off
  * Imports meeting after meeting
  * Until about 80% of max_execution_time is reached
  * Then rewrites import buffer option
  * And refreshes cache

## Settings

Import & Settings page now contains dropdowns for selecting the parent region of a new data source, and for setting the schedule of the invalidation of **all data sources at once**.

Schedule setting ignores actual schedule state and just reschedules on change.

Cron jobs are disabled by default, and can be disabled via the schedule setting.

## Testing

Tested in last few WordPress releases as they went by, but no dedicated testing.

You can use "WP Crontrol" for cron job inspection and scheduling for testing.


mleahy-rns:
Is there any update on this? I'm using the latest version 3.9.3 of the plugin and I don't see any other options for automating data source imports.

tech2serve:
> Is there any update on this? I'm using the latest version 3.9.3 of the plugin and I don't see any other options for automating data source imports.

Unfortunately, no. We've not had the bandwidth nor the expertise to test this PR. When I tried, it was not successful. I reached out to the originator of the PR, but he seems to have disappeared. 

cc: @wokat

kiyote33:
Hi folks,
I'm a retired developer. I might be able to assist in getting this tested and ready for release.  

Let me know...

97eighty:
@kiyote33 

I have this on my list but was waiting until an upcoming milestone was released. Tackle it if you want. I haven't started merging to two.

PR #69, opened 2020-03-15
Temporary closure styling

wokat:
@tech2serve 

See it in action :sweat_smile: already: https://www.alcoholics-anonymous-berlin.de/meetings/daily-reflections-meditation?tsml-day=any&tsml-region=berlin&tsml-type=TC

Tries to deal with this TIAA issue: https://discourse.tiaa-forum.org/t/12-step-meeting-list-plugin-update-temporary-meeting-closure/2434/52

Refactorings:

* Defined simplistic breakpoint-mixins, naming them after the screen-width

Fixes meeting list on small screens:

* Default style of little type-tag is now as it would be on larger devices: simply flows in text behind the meeting name
* Meeting name is not being cut any more, it can break lines as necessary
* Vertical cells on small screen now scale in width as available
* Also fade out meetings of type `Temporary Closure`

Adds highlighting to the `Temporary Closure` type:

* Defines an alert color (currently purplish-red) in `public.scss` only
* Also fades out such meetings on location detail
* Show all `small` tags for that type as `color: $alert_color; font-weight: bold`


tech2serve:
I've tested this with three themes using Safari and a mobile device. Everything looks good so far. Going to test with Chrome and Firefox next. If still good, I'll merge.

If anyone else is testing please holler if you're having any issues.

PR #78, opened 2020-03-26
Basic Online Meetings Support

joshreisner:
* added `conference_url` and `conference_phone` fields to meeting edit page
* added `Online Meeting` type
* can be applied to either meeting or group 
* added ☝️ to API
* added buttons to single meeting template
* crossed out addresses and hid directions buttons when meeting type includes `TC`
* fixed indentation on single meeting template


joshreisner:
Added some CSS to fix overflow issue spotted by @tech2serve 

(should work globally for all frontend buttons)

Screen Shot 2020-03-27 at 8 15 38 AM

PR #81, opened 2020-03-27
Remove tc

brianw-area24:
Add the ability to add or remove the type "Temporary Closure - TC", in bulk


tech2serve:
Wordpress' Bulk Actions allow changes (at least temporary) to meetings coming from a JSON feed. We should either add a conditional warning like this 
 or document clearly that bulk `temporary closure` changes will not be permanent if the meetings are from a JSON feed.

PR #84, opened 2020-03-28
Online meetings frontend styling

tech2serve:
This adds front end, widget styling & single meeting page styling for online meetings. 

To help reduce the clutter next to the meeting name, the temporary closure flag is not displayed if the meeting is also online; instead only "Online meeting" is displayed. The server side logic for this could be cleaned up, but it seems to work so in the interest of speed. We can refactor later.


includes/functions.php
```diff
@@ -423,18 +423,15 @@ function tsml_format_name($name, $types=null) {
 	if (!is_array($types)) $types = array();
 	if (empty($tsml_programs[$tsml_program]['flags']) || !is_array($tsml_programs[$tsml_program]['flags'])) return $name;
 	$append = array();
-	$is_meeting_online = in_array("ONL", $types, TRUE);
+	$meeting_is_online = in_array("ONL", $types, TRUE);
```
joshreisner: better to use single quotes

joshreisner: also the `, TRUE` wouldn't do anything in this case

assets/src/public.js
```diff
@@ -521,13 +521,16 @@ jQuery(function($){
...
-					let isMeetingTempClosedAndOnline = (obj.types.indexOf('ONL') !== -1) && (obj.types.indexOf('TC') !== -1)
-					console.log(obj.name, isMeetingTempClosedAndOnline)
 					//add type 'flags'
 					if (typeof tsml.flags == 'object') {
+						// True if the meeting is temporarily closed, but online option available
+						var meetingIsOnlineAndTC = (obj.types.indexOf('TC') !== -1) && (obj.types.indexOf('ONL') !== -1)
 						for (var i = 0; i < tsml.flags.length; i++) {
-
-							if ((obj.types.indexOf(tsml.flags[i]) !== -1) && (isMeetingTempClosedAndOnline && tsml.flags[i] !== 'TC')) {
+							flagIsTempClosed = tsml.flags[i] === 'TC'
```
joshreisner: prob need a `var` here for consistency

tech2serve: I agree! Not sure why that didn't error?

includes/functions.php
tech2serve: Both changed

PR #87, opened 2020-03-30
Ready for review

tech2serve:
To enable the caller to have the meeting id automatically dialed when calling into a conference, this changes what the server returns so commas and # is left in the number.

Do we still have sufficient validation of phone numbers? We strip all except `0-9,#`.


joshreisner:
Looks 👍 to me!

PR #90, opened 2020-03-31
Fixes JSON import of new online fields

tech2serve:
This should fix the issue with the import of JSON feed not including the `conference_url`, `conference_phone` and `venmo` fields.


joshreisner:
Looks good but @tech2serve I think you need to run `npm run prod`

PR #96, opened 2020-04-04
Tim/add mailing address to export feed

### LINKED ISSUE #76: CVS File vs Main App field mismatch
The CSV  download file has several fields which do not match the fields in the main app. 
I asked WP Support and they sent me here to submit this issue.  https://wordpress.org/support/topic/csv-file-mismatche-with-webpage/#post-12580729

The reason the CSV file is important to us: we can use it as a repository for all our Group/Meeting Info. We can then import into printable templates; keep track of group numbers; keep track of which group is in which district, etc. We don’t have that ability today, it’s currently a kludge of incomplete / scattered data.

I’m not a developer nor an I.T. person, so I truly appreciate all the kind and patient support those of you have provided to Central Offices and volunteer web servants.

Thanks for all you do.
Eric B

tech2serve:
This PR closes #76 AFAIK. I didn't spend a lot of time with the import of sub regions, but I suppose we can go back and look at that. 

Working this, I discovered that we either introduced a bug with groups or it was there and undocumented. I think we introduced it in the last update. When processing the import buffer,  group information was added to the database, but the `group_id` was not added to the appropriate meetings. This has been fixed.

I also reviewed both JSON feed and CSV import/export processes to ensure the new `conference_url` and `conference_phone` fields were properly handled both ways.

In your reviews, please make sure I didn't goof anything with the database, and tell me if we should have done any sanitization of the `mailing_address` field, especially on the import from a csv?


includes/ajax.php
```diff
@@ -476,10 +478,16 @@ function tsml_ajax_import() {
 			}
 			
 			//add custom meeting fields if available
-			foreach (array('types', 'group', 'data_source', 'venmo', 'conference_url', 'conference_phone') as $key) {
+			foreach (array('types', 'data_source', 'conference_url', 'conference_phone') as $key) {
```
joshreisner, line 481: this change should come out; `group` and `venmo` are associated with the group, but this is attaching custom fields to the meeting

includes/ajax.php
```diff
@@ -476,10 +478,16 @@ function tsml_ajax_import() {
...
 				if (!empty($meeting[$key])) add_post_meta($meeting_id, $key, $meeting[$key]);
 			}
 
+			// Add Group Id if applicable
+			if (!empty($meeting['group'])) {
+				$group_id = $groups[$meeting['group']];
+				if (!empty(group_id)) add_post_meta($meeting_id, 'group_id', $group_id);
```
joshreisner, line 488: I'm very puzzled as to how this was missing earlier. Not having this basically means that if you imported from CSV that importing of any group information never worked? I thought it was just two fields that we were focusing on.

tech2serve: I didn't go back and look prior to our last release, but we moved a bunch of fields into an array and looped through it. I thought maybe we introduced that but then?

includes/ajax.php
```diff
@@ -552,6 +568,7 @@ function tsml_ajax_import() {
 				'regions'	=> sprintf(_n('%s region', '%s regions', $regions, '12-step-meeting-list'), number_format_i18n($regions)),
 			),
 		));
+		dd($meetings);
```
joshreisner, line 571: think this needs to come out

includes/functions.php
```diff
@@ -1016,9 +1016,9 @@ function tsml_get_meta($type, $id=null) {
...
+		'tsml_group' => '"website", "website_2", "email", "phone", "mailing_address", "venmo", "last_contact"' . (current_user_can('edit_posts') ? ', "contact_1_name", "contact_1_email", "contact_1_phone", "contact_2_name", "contact_2_email", "contact_2_phone", "contact_3_name", "contact_3_email", "contact_3_phone"' : ''),
 		'tsml_location' => '"formatted_address", "latitude", "longitude"',
-		'tsml_meeting' => '"day", "time", "end_time", "types", "group_id", "website", "website_2", "email", "phone", "last_contact", "conference_url", "conference_phone"' . (current_user_can('edit_posts') ? ', "contact_1_name", "contact_1_email", "contact_1_phone", "contact_2_name", "contact_2_email", "contact_2_phone", "contact_3_name", "contact_3_email", "contact_3_phone"' : ''),
+		'tsml_meeting' => '"day", "time", "end_time", "types", "group_id", "website", "website_2", "email", "phone", "mailing_address", "venmo", "last_contact", "conference_url", "conference_phone"' . (current_user_can('edit_posts') ? ', "contact_1_name", "contact_1_email", "contact_1_phone", "contact_2_name", "contact_2_email", "contact_2_phone", "contact_3_name", "contact_3_email", "contact_3_phone"' : ''),
```
joshreisner, line 1021: think this change needs to come out, since only groups have these fields

tech2serve: Agreed. I should have caught that.

PR #99, opened 2020-04-06
load conference info via ajax to protect it from possible scraping

joshreisner:


joshreisner:
let's not do this anymore :)

PR #104, opened 2020-04-12
Tim/tweak onl reqs and styling bug fixes

### LINKED ISSUE #101: Meeting Type "Women" bug
One of my meetings in version 3.6.4 has "Women" as the type. In the regular meeting list this appears in the groupname: "groupname / Women" and in bold. In the meeting details green it show bold and green. I think this might be a leftover that was missed when we were moving to this latest version. It is not super serious, but it would be better if this feature was removed.

You can see it in action here:
http://al-anon.alateen.on.ca/meetings/?tsml-day=any&tsml-type=W

Thanks for all the hard work - I know changing functionality in a live solution is not easy and there are bound to be bugs. But all in all I feel you are all doing a great job!

Nicky

### LINKED ISSUE #102: Online Meeting type not getting set when importing from JSON
**Describe the bug**
When importing meetings that have conference_url and/or conference_phone via JSON, the Online Meeting type does not get set. Everything else works fine.

**To Reproduce**
Import from a JSON feed that has online meetings. Eg https://www.aasf.co/apps/dbmanager/c360_meeting_json.cfm

**Expected behavior**
The Online Meeting tag should be applied for those meetings.

tech2serve:
Changes:

- Improved public styling so location and address are grayed out in any situation where `TC` is selected as a status/type.
- Added online meeting options for programs other than AA. This should fix styling issues and close #101.
- Added `conference_url` validation for JSON/CSV import.
- Added more checks surrounding `ONL`. 
    - If either `conference_url` or `conference_phone` is considered valid, `ONL` is add/accepted. Closes #102 .
    - If _Online Meeting_ is included in import types, but `conference_url` and `conference_phone` are not valid/empty, the `ONL` type is not added to the meeting.


joshreisner:
Hey @tech2serve I think you want to run `npm run prod` to minify the files

includes/save.php
```diff
@@ -83,6 +84,7 @@ function tsml_save_post($post_id, $post, $update) {
 			array_push($_POST['types'], 'ONL');
 		} 
 	}
+	if (!empty($_POST['conference_phone'])) array_push($_POST['types'], 'ONL');
```
joshreisner, line 87: how about making line 80 

```
if (!empty($_POST['conference_url']) || !empty($_POST['conference_phone'])) {
```

above so ONL doesn't get added twice?

includes/functions.php
```diff
@@ -1260,6 +1260,17 @@ function tsml_import_buffer_set($meetings, $data_source=null) {
...
 		}
 
+		// If Conference URL, validate; or if phone, force 'ONL' type, else remove 'ONL'
+		$meetings[$i]['types'] = array_values(array_diff($meetings[$i]['types'], array('ONL')));
+		if (!empty($meetings[$i]['conference_url'])) {
+			$url = esc_url_raw($meetings[$i]['conference_url'], array('http', 'https'));
+			if (tsml_conference_provider($url)) {
+				$meetings[$i]['conference_url'] = $url;
+				$meetings[$i]['types'][] = 'ONL';
+			} else $meetings[$i]['conference_url'] = null;
```
joshreisner, line 1270: I think it might be more stylistically consistent to write it as 
```
} else {
   $meetings[$i]['conference_url'] = null;
}
```

includes/functions.php
```diff
@@ -1260,6 +1260,17 @@ function tsml_import_buffer_set($meetings, $data_source=null) {
...
+		}
+		if (!empty($meetings[$i]['conference_phone'])) $meetings[$i]['types'][] = 'ONL';
```
joshreisner, line 1272: Might want to make this an `elseif` so that ONL doesn't get added twice

```
} elseif (!empty($meetings[$i]['conference_phone'])) {
  $meetings[$i]['types'][] = 'ONL'; 
}
```

includes/save.php
joshreisner: Mark made me look at this twice and I could be wrong here because I'm not seeing what's on lines 81-83

includes/functions.php
tech2serve: The reason I moved away from `elseif` is because if the `conference_url` isn't empty, but doesn't validate, we still want to check to see if the `conference_phone` will add the `ONL` type.

OSAwebsite:
Not sure this was fixed... have the latest plugin on my staging site and do not see the flag for online meetings, 'women' flag is still there as well....

http://www.iamnow.today/meetings/?tsml-day=any&tsml-query=(elec)+We+Are+Not+Alone+Women%27s+AFG+-+Kingston+(D31)

tech2serve:
@OSAwebsite Please confirm for me what program you have selected for your website.

OSAwebsite:
Program selected is Al-Anon.

PR #106, opened 2020-04-15
Further flag fixes for other programs

tech2serve:
As the title says, I missed adding the `TC` and `ONL` flags for Al-Anon and a couple other programs.

### DIFF
--- a/includes/variables.php
+++ b/includes/variables.php
@@ -500,7 +500,7 @@ function tsml_define_strings() {
 		),
 		'al-anon' => array(
 			'abbr' => __('Al-Anon', '12-step-meeting-list'),
-			'flags' => array('M', 'W'), //for /men and /women at end of meeting name (used in tsml_format_name())
+			'flags' => array('M', 'W', 'TC', 'ONL'), //for /men and /women at end of meeting name (used in tsml_format_name())
 			'name' => __('Al-Anon', '12-step-meeting-list'),
 			'type_descriptions' => array(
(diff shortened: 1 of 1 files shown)


OSAwebsite:
We are almost there... the bold flag for Women is still in place in version 3.6.6
I did update the live site to this version of the plugin, so that it is now in sync with the latest changes but I thought that we were going to take that flag out? You can check it here:
http://al-anon.alateen.on.ca/meetings/?tsml-day=any&tsml-query=women

Both meetings are for women, and they are also online meetings. There is no need to single out the 'women' flag in this way. It is not that important and besides, if we start doing that then we would have to explain why the same does not happen when a meeting is flagged as a 'men' meeting. So, for online meeting and temporarily closed this is a relevant flag, but as far as I am concerned not for anything else...

Thanks for looking into it...

tech2serve:
We had not intended to remove the Men or Women flags from the listings or the single meeting page. We’ve only added the Location Temporarily Closed or the Online Meeting flags as appropriate. We display those alongside Men or Women, whichever type the meeting has assigned by the admin.

I just tested this on my dev site using an Al-Anon program configuration. For one meeting that has the types TC, ONL, and MEN, the flags for Online Meeting and Men are shown. For another meeting in my list that has ONL and WOMEN checked, the Online Meeting and Women flags are showing.

If this is not what you’re experiencing with your site, then I suggest we set up a Zoom call this coming week and let me see your admin page.

PR #111, opened 2020-04-23
Prevent issues with ACF time picker

machouinard:
Hooking tsml_assets() to admin_print_scripts allows us to check $post_type and only enqueue tsml assets for tsml_meeting post type.  Appears to not negatively affect other calls to tsml_assets().

Also:
Prevent PHP undefined index notice.
Use $group->post_title instead of undefined $title for ajax tokens.

### DIFF
--- a/includes/ajax.php
+++ b/includes/ajax.php
@@ -61,7 +61,7 @@ if (!function_exists('tsml_ajax_groups')) {
 				'contact_3_phone'	=> @$group_custom['contact_3_phone'][0],
 				'last_contact'		=> @$group_custom['last_contact'][0],
 				'notes'				=> $group->post_content,
-				'tokens'			=> tsml_string_tokens($title),
+				'tokens'			=> tsml_string_tokens($group->post_title),
 				'type'				=> 'group',
 			);
 		}
--- a/templates/archive-meetings.php
+++ b/templates/archive-meetings.php
@@ -542,7 +542,11 @@ if (($day === null) && !empty($meeting['time'])) {
 									<?php
 break;
 
-            case 'distance': ?>
+            case 'distance':
+            	if ( ! isset( $meeting['distance'] ) ) {
+            		break;
+				}
+            	?>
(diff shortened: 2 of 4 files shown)


includes/ajax.php
```diff
@@ -61,7 +61,7 @@ function tsml_ajax_groups() {
 				'contact_3_phone'	=> @$group_custom['contact_3_phone'][0],
 				'last_contact'		=> @$group_custom['last_contact'][0],
 				'notes'				=> $group->post_content,
-				'tokens'			=> tsml_string_tokens($title),
+				'tokens'			=> tsml_string_tokens($group->post_title),
```
joshreisner, line 64: this looks like it's for a separate issue right?

templates/archive-meetings.php
```diff
@@ -542,7 +542,11 @@ function tsml_body_class($classes)
 									<?php
 break;
 
-            case 'distance': ?>
+            case 'distance':
```
joshreisner, line 545: this seems like it could be for a separate issue. also don't think we should not render the column at all, that could throw off the table's layout

machouinard: Good point.  I saw the display: none and it silence the xdebug notices on my local dev.

joshreisner: Ah, might be quicker to just say echo @$meeting['distance'] below to silence the debug notice

includes/ajax.php
machouinard: Yes, I had some changes in this branch that I'd forgotten about.  Would it be better if I roll back anything not having to do with the ACF issue and push another commit?

joshreisner: Nah, if it needs to be fixed I say let's fix it :)

joshreisner:
Looks good to me, thanks @machouinard !

includes/ajax.php
tech2serve: Cool. his released to any of our other open issues?

Issue #64, opened 2020-03-01
Filtering by district and sub-districts


TheGreenSourcerer:
# Current behavior

Currently, when I select a district on an archive page only the meetings of the groups directly attached to this district are searched.
Is there any way to change this behavior so that all the meetings from sub-districts are displayed as well?

As an example – this list displays no meeting:

https://spis.aa.org.pl/index.php/meetings/?tsml-day=any&tsml-district=region-galicja

# Changed behavior

What I would like to achieve is that is will display all the meetings from intergrupa-kresy AND intergrupa-malopolska AND … AND intergrupa-sadecka which are the sub-districts of the region-galicja district. In other words – select them recursively by the district.

# Proposed solution

WDYT about adding all taxonomy terms - from the top to the bottom of the hierarchy - to the meeting post type?
With this approach, it should be pretty easy to query all the meetings that contain any of the district/sub-district.
The downside is the need for migrating the existing meetings data.

joshreisner:
I could be wrong, but at first glance this seems like a simple bug. 

In the same way that https://spis.aa.org.pl/index.php/meetings/?tsml-day=any&tsml-region=polska includes all the sub-regions, filtering by a district *should* be including its sub-districts, although I see that it is not.

What I'm saying is, I think we need to dig into the filtering to see what's going on, but probably not change the data structure.

TheGreenSourcerer:
Any updates on this issue?

TheGreenSourcerer:
When do you plan to release this fix? 😊

Issue #73, opened 2020-03-19
Filtered meetings dissapeared


TheGreenSourcerer:
**Describe the bug**

After the recent update, when I try to filter the meetings list by district/sub-district/areas, in some cases the list blinks and disappears immediately. When I hit refresh (`CTRL`+`R`/`CMD`+`R`) all the filtered meetings were displayed correctly.

The bug is obviously related to the user filter action.

**To Reproduce**

1. Go to https://spis.aa.org.pl/index.php/meetings/?tsml-day=any&tsml-district=intergrupa-chelmska
2. In "Districts" filter, expand "Region Lublin" and click "Intergrupa Lubelska"
3. The lists blinks and an **empty list is displayed**
4. Hit the browser refresh button (or `CTRL`+`R`/`CMD`+`R`)
5. The list is displayed correctly

The same happens on other sub-districts and areas filters as well.

In some cases, only a part of the list is displayed on changing the filter (and the whole when refreshed).

**Expected behavior**

The list should be displayed with all the meetings assigned to specific sub-district.

**Screenshots**

After changing the sub-district filer:

![image](https://user-images.githubusercontent.com/51343011/77117223-dce29500-6a31-11ea-84e4-e933704737a1.png)

After refreshing the page:

![image](https://user-images.githubusercontent.com/51343011/77117334-19ae8c00-6a32-11ea-8539-90280c9cfa6d.png)

**Desktop (please complete the following information):**

 - OS: Windows 10
 - Browser - observed on Chrome and Edge
 - Version - both were the latest versions

**Smartphone (please complete the following information):**

Was not testes on smartphone

**Additional context**

The network console shows a successful AXAJ call for filtered resources.

After changing the filter, the following JS error is displayed on the browser console:

```
Uncaught TypeError: Cannot read property 'split' of undefined
    at formatAddress (public.min.js?ver=3.5.3:1)
    at Object.<anonymous> (public.min.js?ver=3.5.3:1)
    at Function.each (jquery.js?ver=1.12.4-wp:2)
    at Object.success (public.min.js?ver=3.5.3:1)
    at i (jquery.js?ver=1.12.4-wp:2)
    at Object.fireWith [as resolveWith] (jquery.js?ver=1.12.4-wp:2)
    at x (jquery.js?ver=1.12.4-wp:4)
    at XMLHttpRequest.c (jquery.js?ver=1.12.4-wp:4)
```

Looks like an unhandled null pointer exception.

[...]

PR #120, opened 2020-06-27
selecting a parent district includes children for #64

joshreisner:
changes the logic for districts in filter_meetings.php to match regions


TheGreenSourcerer:
Do I understand correctly, that this is a fix for both those issues:

- https://github.com/code4recovery/12-step-meeting-list/issues/73
- https://github.com/code4recovery/12-step-meeting-list/issues/64

?

joshreisner:
@TheGreenSourcerer yes it's in progress. It still needs to be approved, merged, and published to WP

PR #121, opened 2020-06-27
adding square and paypal as 7th tradition options

joshreisner:
This PR adds Square Cash and PayPal.me as options for 7th Tradition for groups and meetings

Additionally, it adds the capability to have Venmo on meetings (formerly this was groups-only)

Finally, for consistency, it adds this capability to the Mailing Address field as well

Screen Shot 2020-06-27 at 9 52 23 AM


FlipperPA:
Just one thought; do we want to have them input only the the PayPal username rather than the full URL?

We could do something like:

https://www.paypal.me/[joshreisner]

I'm wondering if this would cut down on errors or increase confusion. Thanks!

joshreisner:
@FlipperPA good thought. I'll do that. New commit incoming

joshreisner:
@tech2serve ok this PR should be ready for review now. in addition to doing the above, it should also handle group / meeting contact switching properly.

includes/save.php
```diff
@@ -337,116 +277,63 @@ function tsml_save_post($post_id, $post, $update) {
...
+	//special validation, todo warn user on fail
+	if (!empty($_POST['venmo']) && substr($_POST['venmo'], 0, 1) != '@') {
+		$_POST['venmo'] = null;
+	}
+	if (!empty($_POST['square']) && substr($_POST['square'], 0, 1) != '$') {
+		$_POST['square'] = null;
+	}
+	if (!empty($_POST['paypal']) && strpos($_POST['paypal'], '/') !== false) {
+		$_POST['paypal'] = array_pop(explode('/', $_POST['paypal']));
+	}
```
FlipperPA, line 325: If we're switching this field to just be the paypal.me username, and not the full URL, we'll have to change this. This is still planning on a full URL.

templates/single-meetings.php
```diff
@@ -159,14 +159,23 @@ function tsml_body_class($classes) {
...
-										<?php }?>
+										<?php }
+										if (!empty($meeting->square)) {?>
+											<p class="square">
+												Square Cash: <a href="https://cash.app/<?php echo $meeting->square ?>" target="_blank"><?php echo $meeting->square ?></a>
+											</p>
+										<?php }
+										if (!empty($meeting->paypal)) {?>
+											<p class="paypal">
+												PayPal: <a href="<?php echo $meeting->paypal ?>" target="_blank"><?php echo substr($meeting->paypal, 22) ?></a>
```
FlipperPA, line 175: This "22" also seems to be relying on the PayPal.me URL. If we're sticking with it, can we add a comment to explain the "22" magic number?

tech2serve: At this time neither the `PayPal` URL nor the username get are shown here.

includes/save.php
joshreisner: What this line does is it says if `paypal` contains a `/` then take everything after the last occurrence. The intent was if you entered `https://www.paypal.com/foobar` that it would grab the value `foobar`

templates/single-meetings.php
joshreisner: Yep good catch, this line needs to be fixed

joshreisner:
Hey @tech2serve give it another test, the PayPal info should displaying properly on the meeting detail page and contact info for individual meetings should be showing up in the CSV export

joshreisner:
@tech2serve yes we could have buttons / icons for these payment methods, good idea

joshreisner:
@tech2serve ok here are the buttons. I couldn't find a 'cash' button in Bootstrap 3 so I used Bootstrap 5 buttons instead. Upgraded the online meeting icons to match

Screen Shot 2020-07-01 at 8 35 06 AM

joshreisner:
I've tweaked the appearance of the contact info on the meeting detail page somewhat. This example meeting has a public contact named "Charles." Of course this looks like a lot of buttons, but the great majority of meetings won't have all this stuff filled out.

Screen Shot 2020-07-04 at 10 51 19 AM

joshreisner:
It's up to the local site whether to display contact names, nothing has changed about that. The default is still 'private.' Not sure how this would relate to GDPR.

Ok to merge?

joshreisner:
Great! 

Screen Shot 2020-07-05 at 12 24 16 PM

Issue #119, opened 2020-06-12
Bug: Address gets deleted when moving from Draft to Published


scottvanwa2:
**Describe the bug**
We have moved several of our previously published meetings to "Draft" while we update them with their Virtual Meeting URL's. We then Republish them. The moment we republish the address gets wiped out of the Address field. An address then needs to be reentered and the meeting published again.

**To Reproduce**
1. Go to WP Admin > Meetings > Previously published meeting.
2. Change to type Draft > Update
3.  Click on Publish
4. Address is no longer present.

**Expected behavior**
The address should maintain through draft and publish.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

97eighty:
Hi Scott,
I'm picking this back up. Are you still experiencing problems? Can you tell us which Maps API you are using?

tech2serve:
I just tested with 3.8.0 using Google maps, and did not see an issue. @scottvanwa2, can we close this?

PR #129, opened 2020-07-19
Bw/draft location

brianw-area24:
This PR is to fix issue #119, where an address can get lost when the last meeting at that location gets set to "draft". This happens because in the save function, it doesn't find the tsml_location post when this post also status of "draft". Therefore, what this PR does is sets the status of the parent post of the meeting, which is the tsml_location, to publish. This way, the function finds the correct location, and uses it, thus preserving the address.

At first, it appeared that the problem was in tsml_delete_orphans() function, however, that function seemed to do the right thing. When all the meetings at a location are set to draft, it sets that location to draft. That seems reasonable. So this PR lets the status of the location be draft until the meeting is being saved.

### DIFF
--- a/includes/save.php
+++ b/includes/save.php
@@ -27,12 +27,6 @@ function tsml_save_post($post_id, $post, $update) {
 	//update is always 1, probably because it's actually 'created' when the edit screen first loads (due to autosave)
 	$update = ($post->post_date !== $post->post_modified);
 
-	// Set the meeting's location status to 'publish'
-	// tsml_delete_orphans() will undo this later if necessary
-	if (!empty($post->post_parent)) {
-		wp_update_post(array('ID' => $post->post_parent, 'post_status' => 'publish'));
-	}
-	
 	//sanitize strings (website, website_2, paypal are not included)
 	$strings = array('post_title', 'location', 'formatted_address', 'mailing_address', 'venmo', 'square', 'post_status', 'group', 'last_contact', 'conference_phone');
 	foreach ($strings as $string) {
@@ -200,7 +194,13 @@ function tsml_save_post($post_id, $post, $update) {
 					'post_content'  => $_POST['location_notes'],
 				));
 			}
-	
+
+			// Set the meeting's location status to 'publish'
+			// tsml_delete_orphans() will change it back to draft if necessary
+			if ($locations[0]->post_status != 'publish') {
+				wp_update_post(array('ID' => $location_id, 'post_status' => 'publish'));
+			}
+
 			//latitude longitude only if updated
  			foreach (array('latitude', 'longitude') as $field) {
 				if (!$update || $old_meeting->{$field} != $_POST[$field]) {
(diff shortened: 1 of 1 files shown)


joshreisner:
Hey @brianw-area24 thanks so much for figuring this out. I can't quite figure out how/why locations are getting set to draft status but you're right that's definitely what's happening, and I'm not sure it's a bad thing frankly, I think we do want them as drafts.

So I think there might be an alternate solution then, what do you think of this?

https://github.com/code4recovery/12-step-meeting-list/commit/59d85100d61cbd13a1025e23ffaf9f6f2f1d3311

This alters the location-getting to include locations of any status. That way we're not artificially publishing anything and the location-related code stays together in the same spot in the script.

brianw-area24:
Thanks @joshreisner. I like your suggestion, however that left the tsml_location post with status 'draft', so I deleted what I added yesterday, and instead moved/modified it so it runs the update query only if it needs to.

The tsml_location is getting set to draft in the second half of the function tsml_delete_orphans(). Looks like it's been there for a long time.

I can see 2 side effects of having the tsml_location set to draft.

1. It doesn't add to the # of locations shown on the settings page. That number is the number of published locations
2. Draft locations are not included in the suggestions in the "location" field on the edit screen. For instance, if you add a new meeting, the draft location doesn't come as a suggestion.

I'm not sure if the second one is desired, or if the suggestions should include publish and draft locations. I could go either way on that. I have not dug into where this suggestion list gets populated

joshreisner:
Ah you're right about where it is, thanks.

So big picture: I do think it makes sense to store meeting-less locations as drafts, because we have a permalink for each published location (eg https://aasfmarin.org/locations/marin-lutheran-church) and I don't think that this should be public if there are no meetings there.

Great thinking on these side effect. I think it's good that it wouldn't add to the # of locations on the settings page, just as I hope draft meetings wouldn't show there either. These numbers should align with what the user sees on the front-end.

I do think we want draft locations to show up as suggestions on the edit screen though.

So I'm not sure what the code to set the location to publish does now. Earlier it was to temporarily set it to publish so it was found in the place where you just added `'post_status' => 'any'`, before it was changed back to draft by tsml_delete_orphans. Now what's it doing? Won't it still be set back to draft by tsml_delete_orphans at the end of the script?

brianw-area24:
I knew there was somewhere else location was used, that was a good example.

I did another commit that hopefully will clarify when/why we need to change the location status to publish, ie when the meeting status has changed from draft to publish, we want the location status to change also. In this case, tsml_delete_orphans will not change it back because the meeting is now published.

I will try to look at how to make draft locations show up in the suggestions, however, I think this PR can get included in a release without that, because it's been like that for a long time, and is really a different issue. Depending on how the suggestion dropdown works, I may ask for help with that one :-)

joshreisner:
Ah great, much clearer now, thanks @brianw-area24 

I agree, I'll make a new issue for the suggestion thing

PR #133, opened 2020-07-25
Fixes #116 multiple type filters

### LINKED ISSUE #116: Multiple filter selections
The meeting list front end currently only allows filtering by one criterion, such as "closed" or "online". Member feedback is that it would be very helpful to be able to filter by multiple criteria, such as "closed", "mens", "book study", "wheelchair accessible". I tried using CMD + click and was able to select multiple categories in the dropdown list but the filter actually used only one. Thanks for your consideration!

tech2serve:
1. In `tsml_filter_meetings`, changed constructor so types (string) accurately becomes an array.

1. In `filter_types`, changed logic so all requested types are checked for existence in meeting before including in filtered list.

Closes #116

### DIFF
--- a/includes/filter_meetings.php
+++ b/includes/filter_meetings.php
@@ -81,7 +81,7 @@ class tsml_filter_meetings
         }
 
         if (!empty($arguments['type'])) {
-            $this->type = is_array($arguments['type']) ? array_map('trim', $arguments['type']) : array(trim($arguments['type']));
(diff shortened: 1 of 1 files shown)


joshreisner:
works great for me!

tech2serve:
Great. I'll work on the issues template next (maybe some tomorrow).

PR #141, opened 2020-08-02
Bugfix: Allow Users to Customize Meetings URL

### LINKED ISSUE #140: Can't Customize Meetings URL
In the FAQ, it says that users can change the URL of the meeting list by setting a value for `$tsml_slug` in their theme. This no longer works.

https://wordpress.org/plugins/12-step-meeting-list/#%0Acan%20i%20change%20the%20url%20of%20the%20meetings%20list%3F%0A

In addition, it would be useful for some users to disable the public meeting finder altogether (say if they were using the new shortcode).

joshreisner:
This wil fix #140


joshreisner:
Good catch @tech2serve -- I just added a little code to hide that link to the meeting finder when `$tsml_slug` is null

PR #160, opened 2020-09-05
Fixing jQuery Warning

joshreisner:
the `size()` function has been deprecated, replacing it with the `length` property

jsPrettier went a little crazy reformatting these files, but I think adding some consistency would also be a good thing


joshreisner:
whoops this actually includes the changes in another PR as well, I could take that out if needed but idk how important that is

pjaudiomv (approved):
Yes I like pretty code, I use phpcbf on my php. Large pretty diff looks good to me :)

Issue #146, opened 2020-08-15
Sorting by location is not working as expected


TheGreenSourcerer:
**Describe the bug**

When the user is sorting by the location the results are not ordered properly for Polish character set. 

**To Reproduce**
1. Go to https://spis.aa.org.pl/index.php/meetings/?tsml-day=any&tsml-region=wojewodztwo-mazowieckie
2. Click on 'Lokalizacja' column name (locations)
3. Scroll down to the row with 'Żabia Wola' in location column
4. The results seem to be ordered randomly in several cases

**Expected behavior**

The proper ordering including Polish characters is:

A Ą B C Ć D E Ę F G H I J K L Ł M N Ń O Ó P (Q) R S Ś T U (V) W (X) Y Z Ź Ż

**Screenshots**

![image](https://user-images.githubusercontent.com/51343011/90308926-200f3a00-dee4-11ea-954f-13b963fd3467.png)

![image](https://user-images.githubusercontent.com/51343011/90308991-92801a00-dee4-11ea-9b1b-626a845462b1.png)

![image](https://user-images.githubusercontent.com/51343011/90309021-d3782e80-dee4-11ea-8cd4-9beab564b0e1.png)

**Desktop (please complete the following information):**

 - OS: all
 - Browser: tested on Chrome and Firefox
 - Version: Chrome 84.0.4147.125, Firefox version unknown

**Smartphone (please complete the following information):**

Not tested on smartphone

tech2serve:
Hello @TheGreenSourcerer, I've not been ignoring your issue here, but we didn't get a chance to discuss it among the team members during our last session. We'll try to look closer at this, and we may come back to you asking for some help with the polish alphabet :-)

crispyatl:
It seems the issue is multi-fold. First, the current PHP and Javascript code "sanitize" meeting information strings before using them to sort, and this sanitization process drops some letters from the Polish alphabet (and presumably many other non-latin languages).  For example, "Łochów" becomes "ochow" after sanitization. Second, the sorting itself is done in a locale-naïve way, so that words starting with Z, Ź and Ż would all be sorted as if they were the letter "Z". I'm working on a fix for both of these issues.

TheGreenSourcerer:
[...]

PR #182, opened 2020-10-12
These changes address Issue #146 - Sorting by location is not working…

crispyatl:
… as expected.

In short, sorting works fine for English, but not for many other languages.

Modification Summary

- includes/functions.php:
  - Added tsml_sanitize_data_sort() function.  It is similar to sanitize_title() but it works with UNICODE
- templates/archive-meetings.php:
  - Modified meeting element data-sort attribute to use tsml_sanitize_data_sort() instead of sanitize_title()
- assets/src/public.js:
  - Added sanitizeDataSort() function.  It is similar to sanitizeTitle() but it works with UNICODE
  - Modified meeting element data-sort attribute to use sanitizeDataSort() instead of sanitizeTitle()
  - Modified sortMeetings() to use Intl.Collate() so that sorts are done in a locale-aware fashion.  Previous sort was locale-naïve


joshreisner:
Hey @crispyatl thanks for working on this! @tech2serve would be a better resource here. Thanks

crispyatl:
> Hey @crispyatl thanks for working on this! @tech2serve would be a better resource here. Thanks

Yep - We're already working together on it. Thanks!

tech2serve:
I'm trying to address the conflict with the minified files. I've applied the minified files from `develop` using a suggestion from StackOverflow, but when I tried to push it back, I couldn't use the same ref, and Git ended up creating a new branch. So, now I'm researching.

joshreisner:
this is weird, hope no one minds but I'm going to try to resolve the conflict

tech2serve:
I think I have the steps figured out, and don’t mind doing it.

joshreisner:
@tech2serve sorry didn't see your comment, hope I didn't mess anything up. basically I just pulled down this branch, updated it, and then re-ran `npm run prod` and I think that should take care of it, hopefully

tech2serve:
No problem! Thanks for the assist.

PR #187, opened 2020-10-18
173 allow approx locations

### LINKED ISSUE #173: Better Handling for Approximate Locations
It's possible to use the address field to enter approximate locations. This is great and should be done more and more for online meetings, especially now that Meeting Guide supports this behavior. The problem is that in TSML these approximate locations, eg `Prague, Czechia`: 

1. show as a discrete point on the map
1. show a concrete distance when in 'near me' or 'near location' mode
1. allow a user to get directions

That an address is approximate can be inferred in the geocoding response from Google. For such addresses `[location_type] => APPROXIMATE`

It would be great to also display this info to the user when editing a meeting (eg "Location is approximate, map and directions will be disabled."

Be sure to also add the approximate boolean value to the local geocoding cache.

tech2serve:
**_Testing this will update your database, including adding "3.9.0" as `tsml_version`. The branch will not work if `tsml_version` is already present in `wp_options`_**

NOTE: I may experiment with this more (using `develop`) to see if a more reliable test is available.

*** 
This PR:

1. Tracks if a location is considered approximate (i.e., City and State). All address cache entries will be updated to indicate if the address is approximate. Each location in the database that consists of only City and State will be flagged as approximate.

1. On the admin meeting edit page, the map will not display a marker if the address is approximate.

1. Similarly, the front-end, to include the single meeting and single location pages, will only display a generic map of the approximate location. The map should not display a marker or an information box, nor should the box/button to provide directions be displayed. 

1. The behaviors should be the same regardless if the site is using Google or Mapbox.   

Closes #173


joshreisner:
Definitely working for me on my demo site. Thanks @tech2serve !

joshreisner (approved):

PR #197, opened 2020-10-24
Fix caching issue with tsml_get_meetings()

machouinard:
Looks like a `false` was added around line 905 of includes/functions.php on 3/25/20 with commit 536a8b9c16d6e9 that has been preventing the json meeting cache file from being used.
This is a little refactoring to make sure the cache is used and also ensures the json file is recreated should it somehow get deleted.


joshreisner:
code looks / works great overall @machouinard, ! thanks for fixing the bug I added.

Just one minor thing: I don't see the need for the new `$return` argument on `tsml_get_meetings` -- seems like it's only getting called from `tsml_cache_rebuild` which would work pretty much the same without it. is there something I'm missing?

machouinard:
You're correct @joshreisner .  No idea where I picked up that habit.  It's been removed.

tech2serve:
I'm going to try moving this over to `develop` and will issue it as a beta for additional testing. I don't expect any issues. I'm getting a warning about potentially losing old commit/comments. 

Worst case, I'll need to ask you @machouinard to make a new PR against `develop` vice master. This is a process change we're trying to make.

joshreisner:
thanks! lgtm! 👍

machouinard:
I was wondering about that @tech2serve so I just went with the readme instructions and branched off master.

tech2serve:
No biggie. I do need to get that updated :-)

Issue #208, opened 2020-11-01
Unable to change meeting location with latest 3.9.3 update


marcinyorkpa:
**Describe the bug**
We are running Wordpress 5.5.3 with TSML version 3.9.3

I'm trying to change a location for an existing meeting.

As soon as I change the Address, the "Update" button becomes disabled, and nothing I do will re-enable it. 

I'm following the steps I've always followed for updating a meeting location:

1. Open the meeting
2. Change the address to the new address
3. Give that address a name
4. Attempt to click "Update"

I don't see any errors in the JavaScript console.

I've tried in both Chrome and Firefox and the behavior is the same

**Expected behavior**
The "Update" button is enabled, and submitting the form saves the meeting update

In addition, the "Preview" button does not work as well. It's clickable, but  nothing happens

**Screenshots**

Initial screenshot of button after pulling up the meeting to edit

![Screenshot 2020-11-01 093341](https://user-images.githubusercontent.com/70951029/97805762-0d281e00-1c26-11eb-8ce3-8eacd2d2f796.png)

The existing location

![Screenshot 2020-11-01 093416](https://user-images.githubusercontent.com/70951029/97805770-1d3ffd80-1c26-11eb-84b2-f8a3a746cd2a.png)

Location after editing it. Note that the map update worked as expected

![Screenshot 2020-11-01 093451](https://user-images.githubusercontent.com/70951029/97805784-3183fa80-1c26-11eb-8f39-8f7644139e13.png)

The Update button after scrolling back up to save changes

![Screenshot 2020-11-01 093508](https://user-images.githubusercontent.com/70951029/97805800-4496ca80-1c26-11eb-83cf-b560a578e87f.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome and Firefox
 - Version latest versions of Chrome and Firefox

tech2serve:
Please try these steps and see if it works:

1. Open the meeting.
1. Change the Location name to the new location.
1. Change the address to the new address.
1. Attempt to click "Update".

Put another way, reverse the order or your second and third steps.

marcinyorkpa:
thanks @tech2serve 

I've tried as you suggested, but the behavior persists.

The Update button stays submittable when I change the name, but as soon as I clear the Address field, the button is disabled and never reenables

tech2serve:
> but as soon as I clear the Address field, the button is disabled and never reenables

[...]

Issue #223, opened 2020-11-17
Problem with determing if  location is approximate


AATimes:
In the latest update, almost all our meetings have had the 'is_approximate_location' set to true so the precise map doesn't show up.

I tracked down the problem to be in the function 'decide_if_location_approximate'. It's counting the number of commas in the address returned from google maps and anything less than or equal to 2 sets the flag.

Problem here is Australia is that a valid address returned by google maps has only 2 commas. Standard addresses don't have a comma between the suburb and the state.

1 King St, Richmond, VA 23222, USA
1 King St, Richmond VIC 3121, Australia

How can we fix configure this function to work correctly for us?

tech2serve:
Dang. We knew this was a possibility but I had hoped we'd dodged it.

Do you know, is this true across all of Australia? And what would a real approximate location look like down under? Would it be "Richmond VIC 3121, Australia"?

AATimes:
Yes, this is the standard address format across all Australia set by
Australia Post and followed by google.
Approx addresses have only one comma - "Richmond VIC 3121, Australia"

This works:
if(!function_exists('decide_if_location_approximate')) {
  function decide_if_location_approximate($formatted_address) {
    if (substr_compare($formatted_address, ', Australia', -12))  return
(substr_count($formatted_address, ",") <= 1);
    return (substr_count($formatted_address, ",") <= 2);
  }
}

tech2serve:
You sound like you know your way around php? We're always looking for volunteers to jump in and help maintain the plugin if you're looking for more service :-)

I'm trying to implement this while also addressing a bit of weirdness with how the rest of this feature is (or, is not) working. With luck, I'll have it to beta testers today/tonight, and released shortly thereafter.

AATimes:
Thanks for your prompt response.

I have corrected our data now.

After the plugin update, I think the system uses the flag from google maps
to set the *is_approximate_location *flag rather than relying on
*decide_if_location_**approximate()* to decide. Is that right?

tech2serve:
When new meetings are put in, that is correct. The next release will improve this and I believe help create a more reliable conversion from javascript to php, which I believe may be the cause of some people's problems. 

The `decide_if_location_approximate()` function is used during database updates as we improve this feature. That function will stop being used shortly.

PR #224, opened 2020-11-17
fix: Remove ambiguity associated with `is_approximate_location` in da…

tech2serve:
…tabase

Accomplished through two changes. First is to write `is_approximate_location` for every
location in the database. Second, changed from `add_post_meta` to `update_post_meta` which should
reduce the likelihood of multiple `is_approximate_location` keys in post meta.

see #208

### DIFF
--- a/includes/database_updates.php
+++ b/includes/database_updates.php
@@ -30,11 +30,10 @@ if (!function_exists('db_update_tsml_locations_approximate_location')) {
     // dd($locations);
     foreach ($locations_posts as $location_post) {
       $location_custom = get_post_meta($location_post->ID);
-      if (!array_key_exists('is_approximate_location', $location_custom)) {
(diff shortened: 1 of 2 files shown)


tech2serve:
Changes for how we handle approximate locations are coming in two parts. This first PR makes it so, for now, we're updating approximate location meta data each time the meeting page is loaded, and we now explicitly write every location's `is_approximate_location` with a boolean (1 or blank/null?, apparently). 

The second PR will change the way we test for approximate location in order to handle Australia's case (see #223). Once both PRs are reviewed and merged, I'll change the updater to only run when we bump to version `3.9.4-rc`, which I'll then push to the beta testers.

tech2serve:
I'm killing this as it is no good. I think I'm finding now that the initial javascript-provided boolean is not getting handled properly. I'm experimenting with `json_decode` to see if that solves the form issue.

PR #259, opened 2021-01-05
fix: Check and convert types to string if not already

### LINKED ISSUE #257: ?tsml-day=0 - Choosing Sunday from drop down shows no meetings found
**Describe the bug**
Good afternoon. 

Seeing an issue when choosing SUNDAY from the drop down page, no results are found. Simply refreshing the page then shows the meetings. 

**To Reproduce**
1. Go to https://www.upstateintergroup.org/meetings/
2. Click on DAY - change to SUNDAY
3. Behold, no results
4. REFRESH PAGE - Behold, Results. 

Using Enfold Theme. 
PHP 7.3
WP 5.6 (Just updated today to see if that corrected the issue. It did not. : (
No site cache tools installed

Let me know what else I can do to provide context.

tech2serve:
Some sites have trouble displaying meetings when the 11th Step Meditation
type, being sanitized to create css classes, throws an error because the
11 is an integer instead of a string.

It remains unclear why this only affects a small set of sites.

closes #257


crispyatl:
There are a few issues here:

#1 Instead of:
```
typeClass = typeof obj.types[i] === 'string' ? obj.types[i] : obj.types[i].toString();
classes.push('type-' + sanitizeTitle(typeClass));
```
You can simply do:
`classes.push('type-' + sanitizeTitle(obj.types[i].toString()));`

#2 Instead of fixing this one field, we could fix future issues with any fields by doing the string conversion in sanitizeTitle(). Since the function deals with titles, whatever it is passed should be converted to a string if it isn't already.  I will create a new PR for this.

Issue #255, opened 2021-01-01
MAJOR PROBLEM when refreshing Data Sources under Import & Settings


Area48Webmaster:
When I hit the refresh under Import & Settings, which I do every night, tonight I got this response to all 1,160 meetings and we currently have nothing listed on the website. Please help!

_**Google gave an unexpected response for address 362 State St., Albany, NY, 12210, United States. Response was
(object) array(
   'error_message' => 'This IP, site or mobile application is not authorized to use this API key. Request received from IP address 35.208.14.103, with empty referer',
   'results' => 
  array (
  ),
   'status' => 'REQUEST_DENIED',
)**_

arbliss:
same issue for me as well. doesn't matter whether i supply the plugin with a mapbox api key or a google maps api key.

joshc47:
Area 47 is also experiencing this issue.  We are not nor have ever used Google Maps

TheGreenSourcerer:
[The documentation states](https://wordpress.org/plugins/12-step-meeting-list/#%0Ai%20need%20to%20correct%20a%20meeting%20address%20or%20change%20a%20pin%E2%80%99s%20location%0A) that:

> We get our geocoding positions from Google (this true even if your maps are by Mapbox).

so the error message seems to have much sense.

TheGreenSourcerer:
@tech2serve what's the reason for using a common Google API key for all geocoding requests?
https://github.com/code4recovery/12-step-meeting-list/blob/master/includes/functions.php#L585

tech2serve:
Hey y'all. First off, I do apologize for the inconvenience, but I think you'll understand when I explain this. We have been using a single API key since the very beginning, but something (we're still trying to track down for sure the cause) has caused our geocodes to spike big time, so much so that that little bit of money set aside was exhausted and now there is a several hundred dollar bill from Google. We're not set up to pay this (but we will obviously).

In the meantime, some restrictions were added to the key to ensure we don't end up owing even more money. We are trying to clear this issue quickly. Some of our volunteers are already putting together potential patches.

If you're comfortable with the code, and you can set up your own Google API key, you can change line 585 of the `functions.php` file to your own API key. This will get you up and running. Long term, any update will overwrite this, but that update should contain the fix.

More to follow soon.

joshc47:
Thanks for the update.  Can you let us know how we can donate to help cover your expenses?

arbliss:
[...]

PR #264, opened 2021-01-12
fix: Hot-fix to reduce invalid geocoding calls (errors)

tech2serve:
The `tsml_get_meeting` was called each time a user started
a new meeting even when address values were empty. Because there were no
checks to see if fields were empty, Google was called with an empty address.

A search was made to see if other blocks of code were unexpectedly calling
`tsml_get_meeting`, but none were found.

This commit should help with the issue documented in #255, but only to the
extend that the extra geocodes caused when a meeting is manually being
entered are avoided.


97eighty (approved):

97eighty:
Work correctly for me.

PR #269, opened 2021-01-19
Bw/issue 255

brianw-area24:
This pull request is the beginning using the new API gateway for geocoding. The main things it does is:
- Adds an option on the Import & Settings page so the site can select the type of geocoding they want to do
- Splits out the Google geocoding into it's own function
- Creates a stub of the function that will call the new API gateway

Main thing left to add, is the code in the function to actually call the new API gateway. I need help with how this call works, either someone can add it here, or send me example(s) and I'll try to get it added.


includes/admin_import.php
```diff
@@ -410,6 +410,13 @@ function tmsl_import_page() {
 		delete_option('tsml_mapbox_key');
 	}
 
+	//change geocoding method
```
tech2serve, line 413: In experimenting with this, we should probably validate the choice. For example, I don't have a Google key defined, but was still able to select the option of "Use my Google API key". This should probably alert than a Google API key is needed, and then revert to the previous value?

includes/functions.php
```diff
@@ -681,15 +711,15 @@ function tsml_geocode($address) {
...
-	//cache result
-	$addresses[$address] = $response;
-	update_option('tsml_addresses', $addresses);
-
-	//add a status and return it
-	$response['status'] = 'geocode';
 	return $response;
 }
 
+//function: Call the new api gateway to geocode the address
```
tech2serve, line 726: With the API gateway, I've tried to have it accept `url_encoded` data, and so I'm experimenting with just changing the URI. But, if I can't get that to work, we should pursue this I think.

PR #279, opened 2021-02-04
Added ability to pass an offset value to adjust time filter

97eighty:
The timestamp used to filter a meeting from the meeting list is based on the current_time. When a meetings start time has passed this timestamp it no longer appears. Users have reported that they would like the meeting to continue to display for a period of time longer (i.e. Until after the meeting has ended) than the start of the meeting. This new offset attribute allows a user to adjust the time used in comparison.

[tsml_next_meetings offset="-60"] would allow a meeting to display for an additional 60 minutes beyond current time.


97eighty:
Closing this pull request. Obsolete and not necessary.

PR #348, opened 2021-05-22
Added support for custom mapbox themes

jtwg:
Thank you for the amazing plugin! I'm using it for a site where we want to be able to use Mapbox's dark theme, but the plugin has the Mapbox theme hard-coded. 

This code adds a new global variable, `$tsml_mapbox_theme`, which the user can set in their `functions.php` that the `maps.js` script will make use of if present. I also updated the README with the Mapbox themes which work with the version of the Mapbox library included in this plugin.

The actual changes here are few - I'm not sure why the whitespace is showing as modifications. I did run the build script to generate new `admin.min.js` & `public.min.js` files.


tech2serve:
Thank you! I should be able to check this out in detail this coming week.

PR #407, opened 2021-07-02
updating [tsml_ui] shortcode to use new embed syntax

joshreisner:
(The old syntax was invalid HTML)

https://github.com/code4recovery/tsml-ui/pull/141


joshreisner:
I'm going to update this PR to remove JSDelivr and fast-track it to `master`

PR #424, opened 2021-07-09
fix: Implement URL parsing for 'active' meetings.

### LINKED ISSUE #396: Version: 3.11.0 Issue with "Active" filtering persistence and URL
1. Is it possible to not dim on list view, the Location/Group text if the meeting is active and online? Could you just dim it for inactive meetings?
2. Is it possible to not dim on list view the Region data if the meeting is active and online? Could you just dim it for inactive meetings?
2. Is it possible to have the default be active meetings? I know I can set it up to open as `/meetings/?tsml-day=any&tsml-attendance_option=active`, but when I select back to meeting after viewing a meeting, it reverts back to any type.
3. Would it be possible to leave /women or /men or /online meeting the way it was? That was after the meeting name, in bold green font. It doesn't improve anything to have, for example, (women) on the next line, when the current version shows it in the same line and stands out since it is green color and bold. From our perspective, particularly if it's a men's or women's meeting, it really needs to stand out better and the previous version (3.10.0)  was just fine.

Old Tuesday 6-7

New Tuesday 6-7

Some of these same comments were made on #377 

Thanks.

tech2serve:
v3.11.0 did not parse the URL query for attendance_option='active'
because this is not a pure attendance option. Rather, it is a superset
of online, in_person, hybrid types.

This fix enables admins to properly create a meeting list of only active
meetings (i.e., not inactive), and this choice persists into the user dropdown
for meeting types.

closes #396

### DIFF
--- a/templates/archive-meetings.php
+++ b/templates/archive-meetings.php
@@ -100,8 +100,9 @@ if (!empty($_GET['tsml-type'])) {
 $attendance_options = array();
 if (!empty($_GET['tsml-attendance_option'])) {
     $attendance_option_queries = explode(',', $_GET['tsml-attendance_option']);
+    // $tsml_meeting_attendance_options['active'] = null;
     foreach ($attendance_option_queries as $attendance_option_query) {
-        if (array_key_exists($attendance_option_query, $tsml_meeting_attendance_options)) {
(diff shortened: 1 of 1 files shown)


joshreisner:
this seems like it's on the right track but doesn't yet persist the attendance_option type when you navigate to a meeting and back

brianw-area24 (approved):
I think this does what it is intended to do!

PR #427, opened 2021-07-10
persist options when you navigate to meeting and back

brianw-area24:
Fixed the meeting links, so the search options should persist when you navigate to a meetings details page, then click the "Back to meetings" link. This, hopefully will finish addressing Issue #424 

This PR does change the file assets/src/public.js, however, I did not include `npm run prod` in the PR, because I think @tech2serve is changing CSS, and I didn't want to create any merge conflicts. However, `npm run prod` should be run when the PR is accepted.

### DIFF
--- a/assets/src/public.js
+++ b/assets/src/public.js
@@ -663,12 +663,7 @@ jQuery(function($) {
 										'-' +
 										sort_time +
 										'">' +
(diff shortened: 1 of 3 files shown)


tech2serve (approved):
Thanks, Brian. I'm going to merge this into `develop`, then change the CSS and prepare a release.

Issue #434, opened 2021-07-10
Don't Import Attendance Options


brianw-area24:
**Describe the bug**
Attendance Options are currently being imported in both the CSV and feeds. However, this can cause inconsistencies in the data. It would be better to allow the Plugin to calculate the Attendance Options after the import is completed.

**To Reproduce**
1. Export a CSV file
2. Edit the CSV, and change the Attendance Options column so every meeting is "online"
3. Import the CSV file, deleting all existing meetings
4. Look at the meeting list

**Expected behavior**
When you do this, the plugin shows all meetings as "online", even though not all of them have a conference url or phone. What I would expect to happen would be for the plugin to figure out the attendance options based on the other data that is included in the import file

**Additional context**
This began in version 3.11.0

PR #435, opened 2021-07-10
Csv import

brianw-area24:
This PR fixes the Bug described in Issue #434, where attendance options are imported from a data source, instead of calculated by the plugin. Importing can cause inconsistent meeting information, such as "online" meetings without a URL or Phone number entered. 

Doing it this way allows the Plugin to calculate the correct values after the meetings are imported.


tech2serve (approved):
Thanks!

PR #470, opened 2021-07-25
add link to new pdf service

joshreisner:
this PR adds a link to the new PDF service and adds a little text indicating that the old PDF format will be going away eventually

Screen Shot 2021-07-24 at 6 32 47 PM


joshreisner:
this change is being held back for the next version and will be folded into #482

PR #473, opened 2021-07-26
Performance Fixes

brianw-area24:
Version 3.11.0 introduced a couple pretty major performance issues, especially when the site has a lot of meetings. These were caused because the geocoding method was called for each meeting returned by the tsml_get_meeting and tsml_get_meetings. This cause a lot of unnecessary database calls, which broke sites with a lot of meetings, like Boston.

With this PR, we now store whether an address is approximate (or specific) in the "location" metadata, so we can get that information much more efficiently.

It also fixes an issue when an address is overridden because these address don't include the "approximate" tag. Now, it will assume these addresses are specific addresses because there shouldn't be a need to override an approximate address.

### DIFF
--- a/includes/database_updates.php
+++ b/includes/database_updates.php
@@ -79,13 +79,13 @@ if (!function_exists('tsml_db_set_address_approximate')) {
 
     foreach ($locations as $location) {
       // if location doesn't have the approximate tag
-      if (empty($location['approximate'])) {
-        $location_address = $location['formatted_address'];
+      if (empty($location['address_approximate'])) {
+        $tmp_address = $location['formatted_address'];
 
         // if the cached address has the approximate tag
-        if (!empty($addresses_cache[$location_address]['approximate'])) {
+        if (!empty($addresses_cache[$tmp_address]['approximate'])) {
           // Location in Database doesn't have approximate, and it's in the address cache, so write it to the database
-          update_post_meta($location['location_id'], 'approximate', $addresses_cache[$location_address]['approximate']);
+          update_post_meta($location['location_id'], 'approximate', $addresses_cache[$tmp_address]['approximate']);
         }
       }
     }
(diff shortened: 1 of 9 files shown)


includes/database_updates.php
```diff
@@ -68,3 +68,26 @@ function db_update_remove_all_is_approximate_location_meta() {
...
+
+    foreach ($locations as $location) {
+      // if location doesn't have the approximate tag
+      if (empty($location['address_approximate'])) {
+        $tmp_address = $location['formatted_address'];
+
+        // if the cached address has the approximate tag
+        if (!empty($addresses_cache[$tmp_address]['approximate'])) {
+          // Location in Database doesn't have approximate, and it's in the address cache, so write it to the database
+          update_post_meta($location['location_id'], 'approximate', $addresses_cache[$tmp_address]['approximate']);
```
joshreisner, line 88: hey @brianw-area24 i think this may be setting `$location['approximate']` but checking for `$location['address_approximate']` above

includes/functions.php
```diff
@@ -980,8 +984,11 @@ function tsml_get_meeting($meeting_id=false) {
 			'saddr' => 'Current Location',
 			'q' => $meeting->location,
 		));
+		$meeting->address_approximate = $location->approximate;
```
joshreisner, line 987: for the purpose of the spec, the value that goes into JSON should be `true` or `false` -- could we make this 

```
$meeting->address_approximate = (!empty($location->approximate) && $location->approximate === 'yes');
```

includes/functions.php
```diff
@@ -980,8 +984,11 @@ function tsml_get_meeting($meeting_id=false) {
...
 	}
 
+	if (empty($meeting->address_approximate)) $meeting->address_approximate = 'no';
```
joshreisner, line 990:
```
$meeting->address_approximate = false;
```

includes/functions.php
brianw-area24: Some other code will have to be updated as well, to make it use the approximate field as a boolean, rather than a string of 'yes' or 'no', but doable

joshreisner:
i tested this locally and didn't see any issues. i incremented the version number to test the updates, and everything looks 👍 i think we should go ahead and merge this with `develop`

Issue #393, opened 2021-06-29
Code Cleanup


brianw-area24:
Over time, sometimes parts of the code are no longer used. This issue is a request to start cleaning up some of this old code, which will hopefully make the code a little more readable, and make future updates easier.

I would like to create a new branch, called code_cleanup, for this purpose. Anyone is welcome to add commits to this branch, which should only be used to refactor code to make it cleaner. No new features here.

For example, just a few things I'm aware of include:
- Typo's where function/variable names start with tmsl_ instead of tsml_
- Functions that are no longer used, including db_update_addresses_cache_approximate_location, db_update_tsml_locations_approximate_locations, and decide_if_location_approximate
- A notice about TC being available, that should have been displayed since the end of April, 2020
- Some spacing that makes the code harder to read
- Missing comments

joshreisner:
Complete with release of `3.12`

PR #482, opened 2021-07-31
Code cleanup, PSR-12 standardization, PHP 5.6

### LINKED ISSUE #107: Increase "upcoming" timelimit
Hello,
My idea is: When "upcoming" time is selected in list to display meetings which start at the top of the hour for up to 10 minutes (+/-) into the hour. Let people join meetings in progress.

Thanks!

### LINKED ISSUE #383: chr() expects parameter 1 to be int, string
**Describe the bug**
Warning: chr() expects parameter 1 to be int, string given in /home/q0afnjbk/public_html/wp-content/plugins/12-step-meeting-list/includes/tcpdf/include/tcpdf_fonts.php on line 1671

**To Reproduce**
1. Install it
2. Get District 42's meetings
3. Go to https://aa-stlawrenceny.org/wp-admin/edit.php?post_type=tsml_meeting&page=import and click on full page

**Expected behavior**
No warnings are issued.

I believe using PHP 7.4 has caused these AND MANY OTHER warnings to appear.

I would turn warnings off but for the life of me I cannot figure out how to do so on our hosting service (and besides, warnings are not good to have anyway).

### LINKED ISSUE #481: Code Cleanup Phase II
- i don't think the `if (function_exists(` wrappers on our functions actually do anything. originally i thought this would enable people to override functions from their themes, but plugins are loaded first, so they'll get a critical error if they try to redefine them. i recommend taking them all out
- there are a couple functions in database_upgrades.php that don't have a `tsml_` prefix. it would be good to add this for consistency
- we should include `database_upgrades.php` from `12-step-meeting-list.php` where we do the other includes
- we should use editor extensions `php intelephense` and `prettier` to enforce proper PSR-2 formatting
- looks like WordPress now [requires PHP 5.6.20+](https://wordpress.org/about/requirements/) which means that we can now use anonymous functions and bracket syntax for arrays. i thought this day would never come
- `tsml_ajax_meeting_link` and `tsml_ajax_phone_link` are no longer being used and can safely be removed
- there are a lot of lines commented out - we should clean these out these periodically

joshreisner:
This PR closes #481, #393 by:

* Standardizes our code formatting to PSR-12
* Updates our PHP version requirement going forward to be PHP 5.6 and up (consistent with WordPress)
* Uses anonymous functions to clean up several extra function definitions
* Standardizes our constants and functions to all use the `TSML_` and `tsml_` prefixes
* Removes unused global variables (I recommend PHP Intelephense to spot them, screenshot below)
* Removes unused functions
* Removes old notice about TC being available
* Removes obsolete notice about running php < 5.3

Additionally, it:

* Adds link to new PDF service
* Moves "upcoming" time limit back 10 minutes to fix #107 
* Updates TCPDF to fix #383

Here's an example of an unused variable check:

Screen Shot 2021-07-31 at 11 26 30 AM


pjaudiomv:
Nice I have a precommit hook I use phpcbf from squizlabs https://github.com/squizlabs/PHP_CodeSniffer  I'm a have to check out the ide extensions!

tech2serve:
I have no issues with the proposed approaches: PSR-12 through Intelephense/Prettier, PHP 5.6.

My concern is how to help developers configure their systems to comply with PSR-12. I realize we don't need to address that right now, but it might be useful to do so.

joshreisner:
If there are no objections I'm going to move this into the `develop` branch so we can start Q/A testing it on the demo website.

PR #483, opened 2021-08-01
v3.11.3 Release

tech2serve:
Improve performance
fix Discussion links
fix minor css with meeting pages


joshreisner (approved):
Looks good! I didn't see the minor CSS fix mentioned but wasn't expecting it ;)

includes/functions.php
```diff
@@ -998,6 +998,12 @@ function tsml_get_meeting($meeting_id=false) {
 	$meeting->post_title			= htmlentities($meeting->post_title, ENT_QUOTES);
 	$meeting->notes 				= esc_html($meeting->post_content);
 
+	// Remove TC when online only meeting has approximate address
```
joshreisner, line 1001: i'm not sure why this is needed -- shouldn't these meetings not be TC?

assets/src/public.scss
```diff
@@ -1195,6 +1195,7 @@ body {
     &.attendance-inactive  div.attendance-option {
         color: $color_alert;
     }
+    &.attendance-online.address-specific p.location-address,
```
joshreisner, line 1198: we have a part commented out after this that says
```
    &.tsml-meeting.tsml-type-tc {
        p.location-address,
        address {
            color: #999;
            text-decoration: line-through;
        }
    }
```
wouldn't it have been easier to use that than to add a new `address-specific` class?

includes/functions.php
tech2serve: No. An online meeting that has an approximate address doesn't need TC because we view it as either natively online or it is a meeting that because online and the in-person option went away *permanently*. We only view TC as necessary when we need to tell the user a specific address is not currently available to attend in person.

Unless we've really goofed, that should be consistent with the spec.

joshreisner: Right that's my understanding too - so not sure why we're trying to filter it out on the meeting page here when it shouldn't be in the database at all.

tech2serve: I believe this is because we're cleaning up an issue we introduced in which *all* online meetings initially got tagged as TC. Is that correct @brianw-area24?.

joshreisner: well ok if that's an issue then let's add a database cleanup operation - not saying it needs to be done in this PR, we could create a future ticket like "clean up TC from approximate meetings and remove logic from tsml_get_meeting"

brianw-area24: My suggestion would be to remove lines 1001-1004 now, and leave taking TC out of the database for a future pull release. We should probably do an issue around this...

brianw-area24 (approved):

includes/functions.php
joshreisner: Ok thanks for spotting that Brian. I've removed it now and created this issue: https://github.com/code4recovery/12-step-meeting-list/issues/493

PR #560, opened 2021-10-22
Allow translation

Carlsans:


brianw-area24 (approved):
Yes, this looks good to me!

PR #619, opened 2022-01-01
Add tsml_import_reformat function call - Fixes #439

### LINKED ISSUE #439: Add method for custom/overrideable Import formatting
**Is your feature request related to a problem? **
Need to modify data source records from regional central offices to align with format used by area website. For example, some regional central office websites use sub-regions like 'NorthEast', 'SouthEast', etc. which need to be stripped out as they are meaningless when imported into the area website.

**Describe the solution you'd like**
1 - Create empty function 'tsml_import_reformat' which receives/returns meetings array. Make it overrideable by adding "if (!function_exists('tsml_import_reformat')) before its declaration and give it an action filter so the function can be intercepted and modified from a child theme.
2 - Modify function 'tsml_import_buffer_set' to always pass meetings array through tsml_import_reformat. 

**Describe alternatives you've considered**
Modify functions.php directly, which would of course always get over-written on every version update. 
Or...
1 - Rename function 'tsml_import_reformat_googlesheet' to simply 'tsml_import_reformat',
2 - Add global variable $tsml_domain_identifier with a default  value (i.e. tsml_domain_identifier = 'spreadsheets.google.com') to variables.php, 
3 - Make tsml_import_reformat function overrideable by adding "if (!function_exists('tsml_import_reformat')) before its declaration,
4 - Modify function 'tsml_import_buffer_set' to call tsml_import_reformat instead of tsml_import_reformat_googlesheet, and use the new global variable $tsml_domain_identifier instead of the hard-coded "spreadsheets.google.com". 
5 - Add an action filter for the function so it can be intercepted and modified by the area website code from a child theme.

This is starting to violate the KISS principle. So am going with the top level thoughts...

**Additional context**

97eighty:
Fixes #439

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1288,6 +1288,10 @@ function tsml_import_buffer_set($meetings, $data_source = null)
 		$meetings = tsml_import_reformat_googlesheet($meetings);
 	}
 
+	if (function_exists('tsml_import_reformat')) {
+		$meetings = tsml_import_reformat($meetings);
+	}
+
 	//uppercasing for value matching later
 	$upper_types = array_map('strtoupper', $tsml_programs[$tsml_program]['types']);
 	$upper_days = array_map('strtoupper', $tsml_days);


97eighty:
Duplicate PR - @kiyote33 Fixed this in another branch.

PR #656, opened 2022-01-30
Updating TSML UI URL

joshreisner:
TSML UI has moved from react.meetingguide.org (domain controlled by GSO) to tsml-ui.code4recovery.org (domain controlled by C4R).

Making this change will have no user impact but it will enable me to save considerably on my annual hosting bill. Thanks!


joshreisner:
thank you!

Issue #425, opened 2021-07-09
Contacts are public in JSON even when set to private


joshreisner:
Private contact information is visible in JSON in https://aasanjose.org/wp-admin/admin-ajax.php?action=meetings

Despite "Meeting/Group Contats Are" being set to "Private"

![Screen Shot 2021-07-09 at 7 34 36 AM](https://user-images.githubusercontent.com/1551689/125094687-50c20a00-e088-11eb-82d3-712a17ff80d0.png)

joshreisner:
Hm, this isn't going to be so easy. There have been a lot of refactors to the caching which will require some discussion. Cache rebuilding has been moved to a side effect of `tsml_get_meetings`

brianw-area24:
I've been thinking about this one for a while. I think the way to strip out the contact information would be to put a foreach loop in the `tsml_ajax_meetings()` function. In addition to striping out contact information, we should strip out `attendance_option` as well.

joshreisner:
ah right, `attendance_option` 👍 

the problem with putting it on `tsml_ajax_meetings` is then you have to do that CPU work every time the data is accessed. if you don't write it to JSON at all then you only have to do it when the JSON data is updated

brianw-area24:
Here's what I understand should happen to resolve this issue:
- Meetings cache file should be moved from tsml-cache-<id>.json, to meetings.json
- When writing this cache, use the "Meeting/Group Contact Are" field (from settings) to include/exclude this info in the cache
- The existing admin-ajax.php URLs should continue working for now, but should get deprecated sometime in the future
- CSV export should not use the cache, it should query the database to get all information, including source feed
- CSV import should import all information, including source feed, and if that feed doesn't exist on the imports & settings page, it should create it

PR #718, opened 2022-03-04
Issue425

brianw-area24:
This PR addresses the feed issues from Issue #425.

These include:
- Making the cache file of meetings follow the directive from the "Meeting/Group Contacts Are" (Private/Public) field
- Moving the cache file to /wp-content/meetings.json
- CSV Exports/Imports include all fields, including Meeting/Group contacts, and whether the meeting is from a data_source (feed). This makes it a more complete backup/restore of meetings
- Change the data src URL in the [tsml_ui] short code to use meetings.json

It also fixes a couple of PHP warnings when exporting CSV file when some of the optional fields are empty.

A future PR will address how to handle the "Authorized Apps" section of the settings page. We'll need to have some future discussions about that, and how to deprecate the existing admin-ajax feeds. For now, this PR is a good first step, and can stand on it's own.

### DIFF
--- a/includes/admin_meeting.php
+++ b/includes/admin_meeting.php
@@ -36,6 +36,14 @@ add_action('admin_init', function () {
 		// Delete the attendance_option metadata tag, don't need it
 		delete_metadata('post', 0, 'attendance_option', false, true);
 
+		// Remove old cache info
+		if ($tmpstr = get_option('tsml_cache')) {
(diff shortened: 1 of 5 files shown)


joshreisner (approved):
works great 🥇

97eighty (approved):
As expected.

Issue #688, opened 2022-02-12
Meeting time should not reset to default when day of week changes


kiyote33:
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/652

<div type='discussions-op-text'>

<sup>Originally posted by **OSAwebsite** January 22, 2022</sup>
This has bitten me in the backside more than once and I would like to enter an urgent change request:

Whenever I change the day of a meeting, the times jumps to 12am. This makes no sense to me whatsoever.
Sometimes a group changes their meeting day, but the time usually stays the same. Sometimes the day is given to me in error so I want to be quick and adjust it and then I often miss the time change.

I would like the time, if there is one entered already, to stay the same, even when I change the day. It is very easy to miss and I don't see the reason why it would be changed if all I do is adjust the day....

Pretty please?
Nicky</div>

OSAwebsite:
I just noticed that the latest plugin update still has not fixed this. Please do as soon as possible.
Thanks!

brianw-area24:
I'm working on another issue how, and this is the next one on the list. Hopefully it will make the next release.

OSAwebsite:
Okidoki!


-
Yours in service,
Nicky D
Ontario South Website Coordinator
al-anon.alateen.on.ca

*This too will pass...*

brianw-area24:
Included in v3.14

PR #720, opened 2022-03-04
Keep start time and end time when the day is changed to another day

brianw-area24:
This PR addresses issue #688, where the start/end times change whenever the day is changed.

- When the day is changed from one day to another day, the start/end times don't change
- When a day is changed to "Appointment" the start/end times are erased
- When "Appointment" is changed to a day, the start/end times are set to 00:00/01:00


joshreisner:
works well! i don't think we need to keep the `data-value` properties on these fields anymore, they were there just to store the time if you switch to Appointment and want to switch back. (it seems fine now without this feature.)

PR #722, opened 2022-03-04
TSML UI shortcode compatibility with block themes

joshreisner:
using the TSML UI shortcode in twenty twenty two was not working, it wasn't rendering anything and a warning in the console from TSML UI indicated it couldn't find the div it needs to render.

this PR moves the TSML UI shortcode javascript to be called from the footer, which fixes that.


brianw-area24:
`wp_is_block_theme()` is a newer function, WordPres 5.9.0, I think. Hopefully people are keeping there WP up to date, but just in case, I wonder if we should say `if (function_exists('wp_is_block_theme') && wp_is_block_theme()) {`

This might save a couple support questions. I could also argue that we should get those questions to educate people on the importance of keeping WP up-to-date :-)

joshreisner:
thanks @brianw-area24 that's a good callout, i'll definitely add that!

joshreisner:
i re-marked this PR as a draft since the `get_header` and `get_footer` replacements aren't working yet…

joshreisner:
PR is now ready for review. it just contains the one change to the TSML UI shortcode :)

brianw-area24 (approved):

Issue #368, opened 2021-06-12
Settings Option To Switch User Interface


brianw-area24:
I would like to make it easier for WordPress users to use the new tsml-ui frontend. As an alternative to using a shortcode on a page, I'd like to have a setting on the Import & Settings page where the user could switch back and forth between tsml-ui, and the plugin's legacy interface.

Then on the archive-meetings page, it would check which frontend is chosen. If tsml-ui is chosen, it would simply output the top widget code, next it would output the code generated by the shortcode, and lastly, it would output the bottom widget code.

This would be a good next step in encouraging people to use tsml-ui. For now, the default behavior would be to use the legacy UI, therefore new sites would use legacy for now. However, we may want to change this up sometime in the future...

Grateful-Viv:
If I may add a few comments here, I like the idea of using the TSML UI as the front end once it’s ready, but would like to keep the plug-in front end for an easier way to find and edit meetings for the admin. Using the meeting dashboard is not the easiest way to find the meetings to edit. It can only sort by meeting name, day and time. Yes, can do a search, but it’s so much faster to go to the plug-in front end to find the meeting you’re looking for. 

As far as the widgets you mention, are you suggesting they could work with the UI, such as the html for Meeting Top, Meeting Detail Bottom?

joshreisner:
I think it should default to use the latest TSML UI, but allow the user to select another version, which can be gotten from here:

https://data.jsdelivr.com/v1/package/gh/code4recovery/tsml-ui

default: `https://cdn.jsdelivr.net/gh/code4recovery/tsml-ui/public/app.js`

versioned: `https://cdn.jsdelivr.net/gh/code4recovery/tsml-ui@1.0.1/public/app.js`

brianw-area24:
Yes, I was thinking a selection of 3 options for now, of course the wording could change
1. TSML Plugin - legacy
2. TSML UI - beta
3. TSML UI - stable

joshreisner:
That might be simpler than what I had in mind with listing out the version numbers, but that could work too.

kiyote33:
@joshreisner Do you foresee TSML_UI supporting Google Maps, or is there a reason it can't support Google?

joshreisner:
@kiyote33 i don't think TSML Ui will ever support google maps. adding another map provider would make the javascript bundle too big

kiyote33:
OK. I'll work with that...

kiyote33:
[...]

PR #596, opened 2021-12-19
Widget on the main WordPress Dashboard

brianw-area24:
This pull request adds a widget on the main WordPress Dashboard to highlight things about Code for Recovery is doing. It addresses the initial post of Issue #595, but not the comments which suggest some farther enhancements.

Some initial questions on this include, includes/admin_menu.php an acceptable place to put this code, or should it be somewhere else? Is the wording ok for this static message? And should we wait until we can pull from the website before merging this PR?

PR #745, opened 2022-03-22
Revamp Import & Setting - UI Switching

kiyote33:
Added UI Switch for TSML UI and back to Legacy UI. 
Refactored Import Settings page to group settings and features with tabs and cards. 

Closes Issue #368


includes/variables.php
```diff
@@ -9,7 +9,10 @@
 $tsml_bounds = get_option('tsml_bounds');
 
 //get the secret cache location
-$tsml_cache = '/meetings.json';
+if (!$tsml_cache = get_option('tsml_cache')) {
```
joshreisner, line 12: is this change intentional? or does this overwrite @brianw-area24 's recent change?

readme.txt
```diff
@@ -68,6 +68,20 @@ Note you can add multiple entries to the array below.
...
+* To be safe, always make a backup of your existing meeting list by using the link on the Import tab to export your Meeting List.
+* If you are going to have change detection on multple data sources, you may choose to add the parent organization(s) to your list of Regions first (i.e. District 1, YourCity Intergroup, etc.)
+* Remove the data source (click on the X next to it's Last Refresh timestamp) We suggest first noting the json feed URL (hover over the feed name to view the URL) for use when adding it back 
+* Set data source options: enter a name for your feed, set the feed URL, select the parent region from the Parent Region dropdown, and lastly choose the "Change Detection Enabled" option.
+* Pressing the "Add Data Source" button will register a WordPress Cron Job (tsml_scan_data_source) for the newly added and enabled data source. By default, this cron job is scheduled to run "Once Daily" starting at midnight (12:00 AM). 
+The frequency and scheduled time that the cron job runs is completely configurable by you if the "WP Crontrol" plugin has been installed.
+
+That's it, your done!
```
joshreisner, line 83: typo, should be "you're"

templates/archive-tsml-ui-blocks.php
```diff
@@ -0,0 +1,27 @@
...
+</head>
+<body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
+<div class="wp-site-blocks">
+
+<?php
+echo do_blocks( '<!-- wp:template-part {"slug":"header","theme":"twentytwentytwo","tagName":"header","className":"site-header","layout":{"inherit":true}} /-->' );
+//block_header_area();
+
+echo do_shortcode('[tsml_ui]');
```
joshreisner, line 16: would skip an execution step if it were simply `echo tsml_ui();`

templates/archive-tsml-ui-classic.php
```diff
@@ -0,0 +1,24 @@
...
+	<?php wp_head(); ?>
+</head>
+<body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
+<div class="wp-site-blocks">
+
+<?php
+get_header();
+
+echo do_shortcode('[tsml_ui]');
```
joshreisner, line 15: same here

includes/admin_menu.php
```diff
@@ -39,3 +39,16 @@ function tsml_fix_highlight($parent_file)
...
+add_action('wp_dashboard_setup', 'tsml_dashboard_widgets');
+function tsml_dashboard_widgets()
+{
+    global $wp_meta_boxes;
+
+    wp_add_dashboard_widget('tsml_help_widget', 'Code for Recovery/12 Step Meeting List(TSML) Plugin', 'tsml_dashboard_help', null, null, 'normal', 'high');
+}
+function tsml_dashboard_help()
+{
+    echo '<p><a href="https://code4recovery.org/">Code for Recovery</a> ' . __('is a nonprofit organization of volunteer members building technology services for recovery fellowships, such as AA and Al-Anon. If you need help, please join our discussion forum. If you would like to make a tax-deductible contribution, please', '12-step-meeting-list') . ' <a href="https://code4recovery.org/"> ' . __('visit our website', '12-step-meeting-list') . '.</a></p>';
```
joshreisner, line 53: this isn't a great construction for translations, because it means that "visit our website" has to come at the end of the string, right before the period. as it says in the [documentation](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings), it's better to translate the whole sentence because word order can vary from language to language

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
+			</div>
+
+			<nav class="nav-tab-wrapper" >
+				<a href="?post_type=tsml_meeting&page=import" class="nav-tab <?php if($tab===null):?>nav-tab-active<?php endif; ?>">Import</a>
+				<a href="?post_type=tsml_meeting&page=import&tab=settings" class="nav-tab <?php if($tab==='settings'):?>nav-tab-active<?php endif; ?>">Settings</a>
+				<a href="?post_type=tsml_meeting&page=import&tab=example" class="nav-tab <?php if($tab==='example'):?>nav-tab-active<?php endif; ?>">Spreadsheet Example & Spec</a>
+			</nav>
+
+			<div class="tab-content">
+			<?php switch($tab) :
```
joshreisner, line 644: perhaps these tabs could be in includes? might make this file a bit easier to read. eg `admin_import_[tab].php`

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
+			return;
+		}
+
+		//Get the active tab from the $_GET param
+		$default_tab = null;
+		$tab = isset($_GET['tab']) ? $_GET['tab'] : $default_tab;
+
+		?>
+
+		<style>
```
joshreisner, line 449: would be good to have all css in the css files. in this case, this css belongs in `admin.scss`

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
+			tsml_alert(__('Distance units updated.', '12-step-meeting-list'));
+		}
 
-	//change sharing setting
-	if (!empty($_POST['tsml_sharing']) && isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce)) {
-		$tsml_sharing = ($_POST['tsml_sharing'] == 'open') ? 'open' : 'restricted';
-		update_option('tsml_sharing', $tsml_sharing);
-		tsml_alert(__('Sharing setting updated.', '12-step-meeting-list'));
-	}
+		//change distance units
```
joshreisner, line 272: this should read "change contact display" -- it was wrong before but would be great to fix it here

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
-							<p><?php echo $error ?></p>
-						</div>
-					<?php } elseif ($total = count($meetings)) { ?>
-						<div id="tsml_import_progress" class="progress" data-total="<?php echo $total ?>">
-							<div class="progress-bar"></div>
+		//change user interface
+		if (!empty($_POST['tsml_user_interface']) && isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce)) {
+			$tsml_user_interface = sanitize_text_field($_POST['tsml_user_interface']);
+			update_option('tsml_user_interface', $tsml_user_interface);
+			tsml_alert(__('User Interface Display is now set to <u>' . strtoupper($tsml_user_interface) . '</u>', '12-step-meeting-list') );
```
joshreisner, line 432: would be great to have this be strong instead of underline, since underline implies it's a link. also if it could read *TSML UI* instead of TSML_UI that would be excellent

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
+										<?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
+										<div class="input">
+											<input type="text" name="tsml_add_mapbox_key" placeholder="Enter Mapbox access token here">
+										</div>
+										<div class="btn">
+											<input type="submit" class="button" value="<?php _e('Add', '12-step-meeting-list') ?>">
+										</div>
+									</form>
+
+									<p>*Please note: Only Mapbox is supported by our <b>TSML UI</b> user interface!  Google is not an option.
```
joshreisner, line 604: Might be cleaner to write this: "Please note: TSML UI only supports Mapbox, not Google."

includes/admin_import.php
```diff
@@ -4,743 +4,1212 @@
...
+
+									<p>Be sure to:<br>
+										<span class="dashicons dashicons-yes"></span> Enable the Google Maps Javascript API<br>
+										<span class="dashicons dashicons-yes"></span> Secure your credentials by adding your website URL to the list
+										of allowed referrers
+									</p>
+
+									<p>Once you're done, paste your new key below.</p>
+
+									<form class="columns" method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>" style="margin:15px -5px 22px 0;">
```
joshreisner, line 621: would be good to keep all CSS in CSS files

joshreisner:
sorry for all the feedback! other than the stuff noted above and below, it works well with some light testing locally.

permalink pages should not show legacy ui, they should redirect. eg `/meetings/my-meeting-name` should redirect to `/meetings?meeting=my-meeting-name`. `/locations/my-location` should either redirect to `/meetings` or 404.

might want to use the SVG version of the logo, attaching here:
logo

a few formatting issues: on the import page, at a certain breakpoint the right column wraps over the full-width column. the full-width column goes off the page, and the info bullets aren't showing:
import

on the settings page, more right column breaking strangely. might be nice to have this be 33% up to a certain point and then jump to 50% or 100%.
settings

includes/variables.php
kiyote33: First time I've seen it. I thought I was getting all the most recent changes from develop. Will revert...

readme.txt
kiyote33: Yep

templates/archive-tsml-ui-blocks.php
kiyote33: This is temporary code I borrowed from a wordpress block tutorial ('Creating block templates for custom post types') by Carolina Nymark. It addresses an identified block header part issue which has not been fixed yet the Gutenberg team. There's an obvious problem in that it will only work with the twentytwentytwo theme. I will give your suggestion a shot and see if we can live with the results.

templates/archive-tsml-ui-classic.php
kiyote33: Same...

readme.txt
kiyote33: Fixed

includes/admin_menu.php
kiyote33: Reverted to current develop branch which doesn't have this code.

includes/variables.php
brianw-area24: I think this is a side effect of when changes got pushed up to #596, and then reverted. With the reverts that @kiyote33 has done, I'm hoping we'll be able to merge everything into develop OK, but we'll want to check develop closely when we merge the rest of the PR's for this version.

includes/admin_menu.php
brianw-area24: This is another side effect, because this is code that I had in #596. I did some more reading, and I think I see a better way to handle translation for this paragraph. The html code from the hyperlinks threw me off. I'll push a new commit, hopefully in the next day or two, into that other PR, to hopefully do that translation string better

includes/admin_import.php
kiyote33: Still **TODO**

includes/admin_import.php
kiyote33: Still **TODO**

includes/admin_import.php
kiyote33: Still TODO

kiyote33:
Hi Josh,

 

Thanks for the thorough review. Just a few items to finish off on your list. I don’t quite understand the permalink comment “permalink pages should not show legacy ui, they should redirect.”.  I’m not sure what you are seeing or how to repro it.  A little more help on this one would be appreciated.

 

Wayne

brianw-area24:
I think I can help.  When you have tsml_ui selected, and you click on the meeting details, you get something like:
http://local.test/wp4/meetings/mc-cormick-place-group-8/
This should be redirected to:
http://local.test/wp4/tsml-ui/?meeting=mc-cormick-place-group-8

My suggestion for how to do this, in the `init.php` file, in the function `tsml_single_template()`, check if it's using tsml_ui, and if so, use the `wp_redirect()` function to redirect to this URL

templates/archive-tsml-ui-blocks.php
joshreisner: i was suggesting replacing `echo do_shortcode('[tsml_ui]');` with `echo tsml_ui();` because these lines are equivalent -- sorry if the wrong lines got highlighted.

brianw-area24 (changes requested):

templates/archive-tsml-ui-classic.php
```diff
@@ -0,0 +1,24 @@
...
+<body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
+<div class="wp-site-blocks">
+
+<?php
+get_header();
+
+echo tsml_ui();
+
+get_footer();
```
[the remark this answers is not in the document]
kiyote33: Thanks for your suggestions and help @brianw-area24. They will be in next commit...

joshreisner:
looks a lot better, thanks! there is still some margin strangeness which you may be working on. looks like paragraphs don't have any bottom padding, it would be good to separate these for readability especially in the "where's my info" section. also there's some extra left margin in the description under "import data sources."

templates/archive-tsml-ui-classic.php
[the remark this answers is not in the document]
brianw-area24: Thanks, these changes look great!

brianw-area24 (approved):

includes/admin_import.php
```diff
@@ -4,1000 +4,678 @@
...
+			#import_settings_wrap .import-card {
+				margin: 10px;
+				margin-top: 20px;
+				padding: 0 20px 0 20px;
+				line-height: 100%;
+				height: auto;
+			}
+
+			@media only screen and (min-width: 900px) {
+			.column.one-third { width: 33.2%; }
```
brianw-area24, line 478: With the responsive design, the 3rd column drops down below the first 2 columns pretty quickly when I narrow the browser window (on both Firefox and Opera on the Mac). By narrowing the column width, `.column.one-third { width: 33.2%; }`, from 33.2%, to 33%, I can see all three columns much longer, and the imports & settings page looks much better.

Isn't responsive design fun:-)

kiyote33:
@joshreisner @brianw-area24 @97eighty @tech2serve All the updates are complete for the UI Switching and Import Settings Refactoring. Believe it's ready for 3.14. Please give this PR#745 a final review if you haven't already.

joshreisner:
hey @kiyote33 looks good! i'm seeing a few html issues in the tabs, which i'll mark up.

only functionality issue i see is that i'm still seeing the meeting detail and location detail pages. these should be redirecting as called out above

includes/admin_import_tab.php
```diff
@@ -0,0 +1,240 @@
...
+	</div>
+</div>
+
+<div id="import-1">
+	<div id="col1" class="col one-third import-col">
+		<!-- Put Import CSV section here -->
+		<div id="import-csv" class="postbox" >
+			<div class="inside settings-card">
+				<h1><?php _e('Import CSV', '12-step-meeting-list') ?></h1>
+				<form method="post" style="padding-top: 10px;" action="<?php echo $_SERVER['REQUEST_URI'] ?>" enctype="multipart/form-data">
```
joshreisner, line 145: let's put all CSS in admin.scss

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,333 @@
...
+
+		<div class="postbox">
+		<!-- Put Need Help here -->
+			<div class="inside">
+				<h1><?php _e('Need Help?', '12-step-meeting-list') ?></h1>
+				<div class="settings-card">
+									
+					<p><?php _e("To get information about this product or our organization, simply use one of the linked buttons below. Both the plugin Wiki page and/or the <b>Code For Recovery</b> website are great sources of information. 
+					You can also ask questions directly throght our GitHub discussion forum which is monitored daily by members of our maintenance team.", '12-step-meeting-list') ?></p>
+					<div style="margin:10px;">
```
joshreisner, line 249: CSS

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,333 @@
...
+						<a href="https://github.com/code4recovery/12-step-meeting-list/wiki/" target="_blank" class="button" style=" margin-right: 35px;">
+							<?php _e('Go to our Wiki', '12-step-meeting-list') ?>
+						</a> 
+						<a href="https:///code4recovery.org/" target="_blank" class="button" style="margin-right: 35px;">
+							<?php _e('Code For Recovery website', '12-step-meeting-list') ?>
+						</a>
+						<a href="https://github.com/code4recovery/12-step-meeting-list/discussions" target="_blank" class="button" style="">
```
joshreisner, line 256: style props => admin.scss

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,333 @@
...
+		</div>
+
+		<div class="postbox">
+		<!-- Put Need Help here -->
+			<div class="inside">
+				<h1><?php _e('Need Help?', '12-step-meeting-list') ?></h1>
+				<div class="settings-card">
+									
+					<p><?php _e("To get information about this product or our organization, simply use one of the linked buttons below. Both the plugin Wiki page and/or the <b>Code For Recovery</b> website are great sources of information. 
+					You can also ask questions directly throght our GitHub discussion forum which is monitored daily by members of our maintenance team.", '12-step-meeting-list') ?></p>
```
joshreisner, line 248: there's a typo on the word "through" but i recommend cutting this paragraph, i don't think it adds anything that the user wouldn't surmise from the links themselves.

templates/archive-tsml-ui-classic.php
```diff
@@ -0,0 +1,27 @@
+ <!-- USING TSML UI IN A CLASSIC THEME -->
+<?php
+
```
joshreisner, line 3: need to either call `tsml_assets();` here so that the theme-specific css loads, or else split that out into its own css file and include that.

to see what i mean, check out the page in twenty fourteen.

without styles
before

with styles
after

templates/archive-tsml-ui-classic.php
```diff
@@ -0,0 +1,27 @@
...
+</div>
+<?php if (is_active_sidebar('tsml_meetings_bottom')) { ?>
+    <div class="widgets meetings-widgets meetings-widgets-bottom" role="complementary">
+        <?php dynamic_sidebar('tsml_meetings_bottom') ?>
+    </div>
+<?php } 
+
+get_footer();
+
+?>
```
joshreisner, line 27: it's a best practice not to end your php file with `?>` -- doing so allows whitespace to sneak into the output when sometimes that's not desired

includes/admin_settings_tab.php
kiyote33: 👍

kiyote33:
> hey @kiyote33 looks good! i'm seeing a few html issues in the tabs, which i'll mark up.
> 
> only functionality issue i see is that i'm still seeing the meeting detail and location detail pages. these should be redirecting as called out above

Thought I had this redirect working correctly. It seems to have climbed back in...

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,333 @@
...
+	</div>
+
+	<div id="column3" class="col one-third">
+		<!-- Put About Us here -->
+		<div class="postbox" >
+			<div class="inside">
+				<h1><?php _e('About Us', '12-step-meeting-list') ?></h1>
+				<div class="settings-card">
+					<div class="c4r-logo" style="float:right; margin-top:-55px; margin-right: -10px;" >
+						<a href="https://code4recovery.org"><img src="/wp-content/plugins/12-step-meeting-list/assets/img/code4recovery.svg" alt="Code For Recovery" ></a>
```
tech2serve, line 231: The image display is weird with my theme, currently 2019:

image

includes/admin_import.php
```diff
@@ -4,1000 +4,612 @@
...
+		//change user interface
+		if (!empty($_POST['tsml_user_interface']) && isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce)) {
+			$tsml_user_interface = sanitize_text_field($_POST['tsml_user_interface']);
+			update_option('tsml_user_interface', $tsml_user_interface);
+			if ($tsml_user_interface == 'tsml_ui') {
+				$tsml_ui = "TSML UI";
+			} else {
+				$tsml_ui = "LEGACY UI";
+			}
+			tsml_alert(__('User Interface Display is now set to <strong>' . $tsml_ui . '</strong>', '12-step-meeting-list') );
```
tech2serve, line 437: Let's consider changing the wording to "Meeting list display is now ..."

Reason: The current wording is kind of geeky. :-)

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,333 @@
...
+	</div>
+
+	<div id="column2" class="col one-third "  >
+		<div class="postbox">
+		<!-- Put Switch UI here -->
+			<div class="inside">
+				<h1><?php _e('Switch UI', '12-step-meeting-list') ?></h1>
+				<div class="settings-card" >
+					<h3><strong><?php _e('User Interface Display', '12-step-meeting-list') ?></strong></h3>
+					<p><?php _e('Please select the user interface design that is right for your site. Choose between our latest design that we call <b>TSML UI</b> or stay with the old standard <b>Legacy UI</b>.', '12-step-meeting-list') ?></p>
```
tech2serve, line 132: Instead of "user interface design", how about "meeting listing style"?

templates/archive-tsml-ui-classic.php
kiyote33: Thanks @joshreisner for this tip. It also resolved the problem I was having getting the twentytwentytwo theme to display correctly for both TSML UI and Legacy.

tech2serve (changes requested):
I need to spend a little more time with this when we get a closer to approval/merging. In particular, the meeting list displayed in the Twenty-Nineteen theme is quite ugly after switching to TSML-UI, but frankly I don't know if that is a theme issue, a `tsml_assets()` resolved issue, or something else.

includes/admin_settings_tab.php
kiyote33: Something strange here. The theme should not affect this admin display.

includes/admin_settings_tab.php
kiyote33: Not a fan.  I think most admins would know UI stands for User Interface. I would vote to keep UI...

12-step-meeting-list.php
```diff
@@ -46,6 +46,8 @@ function tsml_import_reformat()
 //include admin files
 if (is_admin()) {
     include TSML_PATH . '/includes/admin_import.php';
+    //include TSML_PATH . '/includes/admin_import_tab.php';
+    //include TSML_PATH . '/includes/admin_settings_tab.php';
```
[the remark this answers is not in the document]
kiyote33: Dead code. I've removed them.

includes/admin_settings_tab.php
kiyote33: This is what I see at the top of the Settings tab with the 2019 theme:
image

includes/admin_import.php
kiyote33: How about "Switch UI is now set to..."?

includes/admin_settings_tab.php
kiyote33: I'm up to 24 themes that I've tested against, and other than some small CSS issues which I was able to resolve with some code in the Additional CSS cutomizer, they all worked reasonable well. Especially now that I've added the tsml_assets() in the template files.

includes/admin_import.php
kiyote33: @tech2serve I've gone with the "Switch UI is now set to..." for now. Will change if this is important to you, as I'm not totally against your suggestion.

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,337 @@
...
+
+	<div id="column3" class="col one-third">
+		<!-- Put About Us here -->
+		<div class="postbox" >
+			<div class="inside">
+				<h1><?php _e('About Us', '12-step-meeting-list') ?></h1>
+				<div class="settings-card">
+					<div>
+						<p id="c4r-logo">
+							<img src="/wp-content/plugins/12-step-meeting-list/assets/img/code4recovery.svg" alt="Code For Recovery" />
```
joshreisner, line 232: we should be using a variable such as `WP_PLUGIN_DIR` for this path. this image will be broken on sites that don't have wp-content in their root directory.

includes/admin_settings_tab.php
```diff
@@ -0,0 +1,337 @@
...
+							<?php _e('This <b>12 Step Meeting List</b> plugin (TSML) is one of the free services offered by the nonprofit organization <b>Code For Recovery</b> whose volunteer members build and maintain technology services for recovery fellowships such as AA and Al-Anon.', 
+							'12-step-meeting-list') ?>
+						</p>										
+					</div>
+
+					<div  style="float: left;">
+																
+					</div>
```
joshreisner, line 240: 🔪 looks like dead code

includes/admin_settings_tab.php
joshreisner: actually i think we should do this the way we do in `tsml_assets` which is to use `plugins_url('../assets/img/code4recovery.svg', __FILE__)`

includes/admin_import.php
tech2serve: I'd like to see @brianw-area24 or @97eighty weigh in. I note that in no place do we say explicitly that the UI change only affects the public facing meeting list.

includes/init.php
```diff
@@ -28,13 +38,30 @@ function tsml_single_template($template)
         global $post;
         if ($post->post_type == 'tsml_meeting') {
             $user_theme_file = get_stylesheet_directory() . '/single-meetings.php';
+           
             if (file_exists($user_theme_file)) {
-                return $user_theme_file;
+                 if ($tsml_user_interface == 'tsml_ui') { 
+                    wp_redirect(add_query_arg('tsml_ui', $tsml_user_interface));
```
joshreisner, line 44: this still isn't working. i did some digging and there are a few things going on here:

* `$tsml_user_interface` needs to be added to `global` above
* this shouldn't be inside the `if (file_exists` statement, because that indicates if a user has supplied their own meeting detail page template, which is immaterial here
* it redirects to the wrong place. it's just adding `?tsml_ui=tsml_ui` to the URL. we want to redirect to the meetings page (`get_post_type_archive_link('tsml_meeting')`) and append `&` or `?` depending on if there's a query string, then `meeting=` and the meeting slug

includes/init.php
```diff
@@ -25,16 +35,30 @@ function tsml_archive_template($template)
...
             $user_theme_file = get_stylesheet_directory() . '/single-locations.php';
+            
+            if ($tsml_user_interface == 'tsml_ui') { 
+                $mtg_permalink = get_post_type_archive_link('tsml_meeting');
+                wp_redirect(add_query_arg('meeting', $mtg_permalink ));
+            }
+            
+            if (file_exists($user_theme_file)) {
+                return $user_theme_file;
+            }
```
joshreisner, line 60: there are two of these one after the next. i believe one of these belongs in the if statement above. otherwise we'd be removing the ability to override the meetings detail page

includes/init.php
```diff
@@ -25,16 +35,30 @@ function tsml_archive_template($template)
...
+        global $post, $tsml_user_interface;
         if ($post->post_type == 'tsml_meeting') {
             $user_theme_file = get_stylesheet_directory() . '/single-meetings.php';
-            if (file_exists($user_theme_file)) {
-                return $user_theme_file;
+             
+            if ($tsml_user_interface == 'tsml_ui') { 
+                    // output we are looking for: https://gutenberg.dev.cc/meetings?meeting=never-stand-alone-group  or  https://gutenberg.dev.cc/meetings?meeting=807-1 
+                $mtg_permalink = get_post_type_archive_link('tsml_meeting');
+                wp_redirect(add_query_arg('meeting', $mtg_permalink ));
```
joshreisner, line 45: this isn't working still, it starts a redirect loop

includes/init.php
```diff
@@ -25,16 +35,30 @@ function tsml_archive_template($template)
...
             }
+            
 
             return dirname(__FILE__) . '/../templates/single-meetings.php';
         } elseif ($post->post_type == 'tsml_location') {
             $user_theme_file = get_stylesheet_directory() . '/single-locations.php';
+            
+            if ($tsml_user_interface == 'tsml_ui') { 
+                $mtg_permalink = get_post_type_archive_link('tsml_meeting');
+                wp_redirect(add_query_arg('meeting', $mtg_permalink ));
```
joshreisner, line 55: try updating this to 
```
wp_redirect($mtg_permalink);
exit;
```

readme.txt
```diff
@@ -68,6 +68,32 @@ Note you can add multiple entries to the array below.
...
+To avoid this warning and prevent a refresh from altering an edited record it's necessary to follow a few simple steps to reimport the data source records:
+
+1 Make a backup of your existing meeting list by using the export link found on the Import tab of the Import & Settings page.
+1 Open the exported file (meetings.csv) which you should find in your local Downloads folder.
+1 Delete the entire 'Data Source' column found near the far right and then Save the file (recommend using Save As to rename the file to something unique such as my-meetings.csv).
+1 Remove the imported data source (click on the X next to it's Last Refresh timestamp). 
+1 Import the saved file using the Import CSV feature on the Import & Settings page. 
+
+Your meeting list records will now no longer display a warning message when being edited, and will not be overwritten by a data source refresh operation!
+
```
joshreisner, line 96: i think we should discard changes to this file since this it's not relevant to the PR

templates/archive-meetings.php
```diff
@@ -668,4 +673,12 @@ public function end_el(&$output, $item, $depth = 0, $args = [])
 </script>
 
 <?php
-get_footer();
+if (function_exists('wp_is_block_theme') && wp_is_block_theme()) {
+   echo do_blocks('<!-- wp:template-part {"slug":"footer","theme":"twentytwentytwo","tagName":"footer","className":"site-footer","layout":{"inherit":true}} /-->');
```
joshreisner, line 677: looks like this adds functionality specific to the twenty twenty two theme. what happens if it's a block theme but not twenty twenty two? should we be checking if that theme is installed? or can we think of a solution for block themes that doesn't require twenty twenty two?

templates/archive-tsml-ui-blocks.php
```diff
@@ -0,0 +1,41 @@
...
+	<?php wp_head(); ?>
+</head>
+<body <?php body_class(); ?>>
+<?php wp_body_open(); ?>
+<div class="wp-site-blocks">
+
+<?php
+tsml_assets();
+
+echo do_blocks( '<!-- wp:template-part {"slug":"header","theme":"twentytwentytwo","tagName":"header","className":"site-header","layout":{"inherit":true}} /-->' );
```
joshreisner, line 15: also here: looks like this adds functionality specific to the twenty twenty two theme. what happens if it's a block theme but not twenty twenty two? should we be checking if that theme is installed? or can we think of a solution for block themes that doesn't require twenty twenty two?

joshreisner:
when the user makes a change to any of the cards, the confirmation message shows up inside the "General" card -- this seems weird

admin-message

assets/src/admin.scss
```diff
@@ -4,29 +4,37 @@
 
 body {
 	//group Information (optional) gray
-	h2 small, h3 small { 
+	h2 small, h3 small {
```
joshreisner, line 7: looks like most of the changes here are from prettier (which is great) but i wonder if we could commit the whitespace changes to `develop` first and then that would help us see just the lines changed?

readme.txt
kiyote33: You're right, it's not relevant. Just trying to sneak it in when I realized I missed it on the last PR for change detection. No real reason for different list types. I put the help topics together at different times and wanted to see how numbering rendered.  Bullets are probably fine.

templates/archive-meetings.php
kiyote33: My understanding is that the requirement for a unique theme name in the template-part function call is considered a bug in the 5.9 code base, and that a fix is expected from the Gutenberg team. Just don't know when that will be at this point. We will likely need to address when a fix comes out.

kiyote33:
@brianw-area24 @97eighty @tech2serve @joshreisner Seems like we may be ready for a merge. Can we get everyone signed off before doing so?

joshreisner:
@kiyote33 thanks, it's looking good. the only is i see at this point is the block-theme stuff does not seem ready to me. per my comment above, we are assuming that twenty twenty two is installed if we see a block theme. if you install a different one (i tried one called "skatepark" you see no header or footer, and the html is not valid. 

also i don't think it really fixes twenty twenty two either, see screenshot below.

i'd rather not introduce code that only partly fixes one theme and causes a regression in others. my recommendation is to remove `wp_is_block_theme` checks in `archive-meetings.php` and `init.php`, remove `archive-tsml-ui-blocks.php` and rename `archive-tsml-ui-classic.php` to `archive-tsml-ui.php`. 

twenty twenty two: 
twenty-twenty-two

joshreisner:
but it's fine with me if you want to merge this PR and we as a team can handle it in a future PR. all i'm saying here is i don't think we should ship this block theme support in `3.14` unless we confirm it works right in twenty twenty two and other block themes first…

kiyote33:
That's fine. It definitely does not work right at this time, so I will follow your suggestions.

joshreisner (approved):
🎉 🥇

PR #769, opened 2022-04-15
V3.14.1

brianw-area24:
Prep for version 3.14.1

- Removes /wp-content/meetings.json file if it exists
- Changes the way the data file url is generated in [tsml_ui] shortcode


kiyote33 (approved):
Everything seems to work correctly in the multisite environment that I spun up for this review. 3 subdomains loaded with their own data, and the parent site pulling data from all 3.  All 4 sites displayed meeting information correctly as expected, although the caching file name (**meetings-cache-1.json**) was the same on all sites.

kiyote33 (approved):
Very nice! No problems detected in Test. I think we are good to go.

Issue #759, opened 2022-04-13
Expand change detection email report


kiyote33:
### Requested Feature/Enhancment

After matching feed records with db records, list each unmatched db records as a removed meeting (i.e. My AA Group removed: last-refresh-timestamp). Conversely, all recently updated feed records should be matched with db records and  listed as changed meetings, and the unmatched feed records as newly added meetings (i.e. My New AA Group added: last-update-timestamp).

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

kiyote33:
Here is a view of the output from my proposed enhancement. Any thoughts or comments?
![image](https://user-images.githubusercontent.com/57125427/163912613-bbe2facb-8bba-4f8d-bd86-28cb8838e095.png)

joshreisner:
nice! maybe add a link to the import page so the user has some place to click?

kiyote33:
Implemented in 3.14.3

PR #776, opened 2022-04-20
Wp/expand email report

kiyote33:
Issue #759 ready for review


joshreisner:
i think the base should be develop, right?

PR #777, opened 2022-04-20
Wp#issue759

kiyote33:
Expand Change Detection Email Report

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2002,7 +2002,7 @@ if (!function_exists('tsml_scan_data_source')) {
 					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
 					$message .= implode('', $meetings_updated);	
 					$message .= "</tbody></table><br>";
-					$import_page_url = 'https://' . admin_url() . '/wp-admin/edit.php?post_type=tsml_meeting&page=import';
+					$import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');
 					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Settings page</a>";
 
 					// send Changes Detected email
@@ -2081,8 +2081,7 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 	// list changed and new meetings found in the data source feed
 	foreach ($feed_meetings as $meeting) {
 	
-		$dayofweek = $dow_number = '';
-		tsml_get_day_of_week_info( $meeting['day'], $dayofweek, $dow_number);
+		list($dayofweek, $dow_number) = tsml_get_day_of_week_info($meeting['day']);
 	    $meeting_slug =  $meeting['slug'];
 
 		// numeric slugs may need some reformatting
@@ -2094,14 +2093,13 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 	    $is_matched = in_array( $meeting_slug, $db_slugs );
 
 		// add slug to feed array to help determine current db removals later on...
-		$feed_slugs[$meeting['slug']] = $meeting_slug;
+		$feed_slugs[] = $meeting_slug;
 
 		// has the meeting been updated since the last refresh?
 		$current_meeting_last_update = strtotime($meeting['updated']);
 		if ($current_meeting_last_update > $data_source_last_refresh) {
-
 			$meeting_name = $meeting['name'];
-			$meeting_update_date = date("M j, Y  g:i a", $current_meeting_last_update);
+			$meeting_update_date = date('M j, Y  g:i a', $current_meeting_last_update);
 			
 			if ($is_matched) {
 				$message_lines[] = "<tr style='color:gray;'><td>Change</td><td >$meeting_name</td><td>$dayofweek</td><td>$meeting_update_date</td></tr>";
@@ -2115,14 +2113,13 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 	// list meetings in local database which are not matched with feed
 	foreach ($db_meetings as $db_meeting) {
 
-		$dayofweek = $dow_number = '';
-		tsml_get_day_of_week_info( $db_meeting['day'], $dayofweek, $dow_number);
+		list($dayofweek, $dow_number) = tsml_get_day_of_week_info($db_meeting['day']);
 	    $meeting_slug = $db_meeting['slug'];
 
 	    $is_matched = in_array( $meeting_slug, $feed_slugs );
 
 		if (!$is_matched) {
-			$meeting_update_date = date("M j, Y  g:i a", $data_source_last_refresh);
+			$meeting_update_date = date('M j, Y  g:i a', $data_source_last_refresh);
 			$meeting_name = $db_meeting['name'];
 			$message_lines[] = "<tr style='color:red;'><td>Remove</td><td >$meeting_name</td><td>$dayofweek</td><td>* $meeting_update_date</td></tr>";
 		}
@@ -2133,14 +2130,13 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 
 //function:	Returns corresponding day of week string and number for the day input
 if (!function_exists('tsml_get_day_of_week_info')) {
-	function tsml_get_day_of_week_info( $meeting_day_input, &$dayofweek, &$dow_number )
+	function tsml_get_day_of_week_info( $meeting_day_input, $dayofweek = '', $dow_number = '' )
 	{
 		global $tsml_days;
 		
-		// when day is like "Monday" convert to number
+		// when day is like "Sunday" convert to number 0
 		if ( in_array($meeting_day_input, $tsml_days) ) {
-			$dow_number = date('N', strtotime( $meeting_day_input ) );
-			if ($meeting_day_input == 'Sunday') {$dow_number = '0';}
+			$dow_number = array_search($meeting_day_input, $tsml_days);
 
 		} elseif ( is_array($meeting_day_input) ) {
 			$dow_number = implode("",  $meeting_day_input);
@@ -2150,11 +2146,9 @@ if (!function_exists('tsml_get_day_of_week_info')) {
 		}
 
 		// only accept valid day of week numbers
-		if (($dow_number >= 0) && ($dow_number <= 6)) {
-		    $dayofweek = $tsml_days[$dow_number];
-		} else {
-		    $dayofweek = 'invalid value';
-		}
+		$dayofweek = array_key_exists($dow_number, $tsml_days) ? $tsml_days[$dow_number] : 'invalid value';
+
+		return [$dayofweek, $dow_number];
 	}
 }
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1986,7 +1986,7 @@ if (!function_exists('tsml_scan_data_source')) {
 
 				// check import feed for changes and return list summing up changes detected
 				$meetings_updated = tsml_import_changes($meetings, $data_source_url, $data_source_last_import);
-				
+
 				if (count($meetings_updated) > 0) {
 					// Send Email notifying Admins that this Data Source needs updating
 					$message = "Data Source changes were detected during a scheduled sychronization check with this feed: $data_source_url. Your website meeting list details based on the $data_source_name feed are no longer in sync. <br><br>Please sign-in to your website and refresh the $data_source_name Data Source feed found on the Meetings Import & Settings page.<br><br>";
@@ -2000,7 +2000,7 @@ if (!function_exists('tsml_scan_data_source')) {
 					$message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
 					$message .= "<br><b><u>Detected Difference</b></u><br>";
 					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
-					$message .= implode('', $meetings_updated);	
+					$message .= implode('', $meetings_updated);
 					$message .= "</tbody></table><br>";
 					$import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');
 					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Settings page</a>";
@@ -2068,7 +2068,7 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 		__('Thursday', '12-step-meeting-list'),
 		__('Friday', '12-step-meeting-list'),
 		__('Saturday', '12-step-meeting-list'),
-	];	
+	];
 
 	// get local meetings 
 	$all_db_meetings = tsml_get_meetings();
@@ -2077,9 +2077,9 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 
 	/* filter out all but the data source meetings  */
 	foreach ($all_db_meetings as $db_meeting) {
-	    $db_id = $db_meeting['id'];
+		$db_id = $db_meeting['id'];
 		if (in_array($db_id, $ds_ids)) {
-		    array_push($db_meetings, $db_meeting);
+			array_push($db_meetings, $db_meeting);
 		}
 	}
 
@@ -2089,17 +2089,17 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 
 	// list changed and new meetings found in the data source feed
 	foreach ($feed_meetings as $meeting) {
-	
+
 		list($day_of_week, $dow_number) = tsml_get_day_of_week_info($meeting['day'], $week_days);
-	    $meeting_slug =  $meeting['slug'];
+		$meeting_slug =  $meeting['slug'];
 
 		// numeric slugs may need some reformatting
-		if ( is_numeric( $meeting_slug )) {
-			$meeting_slug .= '-' . $dow_number; 
+		if (is_numeric($meeting_slug)) {
+			$meeting_slug .= '-' . $dow_number;
 		}
 
 		// match feed/database on unique slug
-	    $is_matched = in_array( $meeting_slug, $db_slugs );
+		$is_matched = in_array($meeting_slug, $db_slugs);
 
 		// add slug to feed array to help determine current db removals later on...
 		$feed_slugs[] = $meeting_slug;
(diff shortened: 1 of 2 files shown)


includes/functions.php
```diff
@@ -1985,26 +1985,28 @@ function tsml_scan_data_source($data_source_url)
...
-						}
+					$message .= "data source feed cnt: $feedCount<br>";
+					$message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
+					$message .= "<br><b><u>Detected Difference</b></u><br>";
+					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
+					foreach	($meetings_updated as $key => $value) {
+						$message .= $value;
 					}
+					$message .= "</tbody></table><br>";
+					$import_page_url = 'https://' . $_SERVER['SERVER_NAME'] . '/wp-admin/edit.php?post_type=tsml_meeting&page=import';
```
joshreisner, line 2008: not all wordpress sites are at the root level. best to use a function like `admin_url()` to find the proper address.

includes/functions.php
```diff
@@ -1985,26 +1985,28 @@ function tsml_scan_data_source($data_source_url)
...
+					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: table;font-size: 18px;'>Goto Import & Settings page...</a>";
```
joshreisner, line 2009: "Go to Import & Settings page" (no ellipsis, space between go and to)

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
 	}
 }
 
-//function:	Returns boolean indicator when data source changes detected
-function tsml_import_has_changes($meetings, $data_source_count_meetings, $data_source_last_refresh)
+//function:	Returns summary list of modified records when data source changes detected
+function tsml_import_has_changes($data_source_url, $meetings, $data_source_count_meetings, $data_source_last_refresh)
```
joshreisner, line 2064: this function name should reflect its new purpose. perhaps `tsml_import_changes`? also `$data_source_count_meetings` is not used and can be removed

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
 {
+	$is_there_change = false;
+	$meetings_updated = $db_meetings = $feed_slugs = $message_lines = [];
+	$days = [0 => 'Sunday',1 => 'Monday',2 => 'Tuesday',3 => 'Wednesday',4 => 'Thursday',5 => 'Friday',6 => 'Saturday'];
```
joshreisner, line 2068: can use `$tsml_days` here

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+	    $db_id = $db_meeting['id'];
+		if (in_array($db_id, $ds_ids)) {
+		    array_push($db_meetings, $db_meeting);
+		}
+	}
+
+	// create array of db slugs for matching
+	$db_slugs = array_column($db_meetings, 'slug', 'id');
+	sort($db_slugs);
+	$db_slugs_cnt = count($db_slugs);
```
joshreisner, line 2086: this variable is not used and can be removed

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+	// list changed and new meetings found in the ds feed
+	foreach ($meetings as $meeting) {
 
-		if (count($meetings) > $data_source_count_meetings) {
-			$meetings_updated[] = 'Feed Records: ' . count($meetings) . ' more than DB Records: ' . $data_source_count_meetings;
-		} elseif (count($meetings) < $data_source_count_meetings) {
-			$meetings_updated[] = 'Feed Records: ' . count($meetings) . ' less than DB Records: ' . $data_source_count_meetings;
+	    $mtg_slug =  $meeting['slug'];
+		if (is_array($meeting['day'])) {
+			$dow = implode("",  $meeting['day']);
```
joshreisner, line 2093: what happens if `$meeting['day']` is an array, and $dow becomes, say the string `24`? Does it show up like that in the report? feel like this could be refactored a little to support "Tuesday, Thursday"

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+			if ($is_matched) {
+				$message_lines[] = "<tr style='color:gray;'><td>Change</td><td >$mtg_name</td><td>$dayofweek</td><td>$mtg_updte</td></tr>";
+			}
+			else {
+				$message_lines[] = "<tr style='color:green;'><td>Add New</td><td >$mtg_name</td><td>$dayofweek</td><td>$mtg_updte</td></tr>";
+			}
+		}
+	}
+
+	sort($feed_slugs);
```
joshreisner, line 2130: why sort `$feed_slugs` here? doesn't look like it's being used

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+			}
+			$updated = $data_source_last_refresh;
+			$cur_meeting_lastupdate = strtotime($updated);
+			$mtg_updte = date("M j, Y  g:i a", $updated);
+			$mtg_name = $db_meeting['name'];
+			$message_lines[] = "<tr style='color:red;'><td>Remove</td><td >$mtg_name</td><td>$dayofweek</td><td>* $mtg_updte</td></tr>";
+		}
+	}
+
+	if ($is_there_change) {
```
joshreisner, line 2159: i think we could remove the `$is_there_change` and `$meetings_updated` variables from this function and simply return `$message_lines` without altering the function's behavior

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+	    $is_matched = in_array( $mtg_slug, $feed_slugs );
+
+		if (!$is_matched) {
+		    $is_there_change = true;
+			if ( is_numeric( $dow )) {
+				$dayofweek = $days[date($dow)]; 
+			} else {
+				$dayofweek = $dow; 
+			}
+			$updated = $data_source_last_refresh;
```
joshreisner, line 2151: not sure why this variable needs to be created, couldn't we just use `$data_source_last_refresh` below?

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+			$cur_meeting_lastupdate = strtotime($updated);
```
joshreisner, line 2152: i would love it if we could be a little more intentional about variable names. i think that aside from obvious shortenings like `db` or `id` that we should try to spell out words. also snake casing should be consistently applied between words. so this could be `$current_meeting_last_update` and `$mtg_updte` could be `$meeting_update`.

this will make for a cleaner appearance and will be much easier for developers who don't have english as a native language.

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
 		// has meeting been updated?
 		$updated = $meeting['updated'];
 		$cur_meeting_lastupdate = strtotime($updated);
-
 		if ($cur_meeting_lastupdate > $data_source_last_refresh) {
+			$is_there_change = true;
 			$mtg_name = $meeting['name'];
-			$mtg_updte = date("l F j, Y  h:i a", $cur_meeting_lastupdate);
-			$meetings_updated[] = "$mtg_name updated: $mtg_updte";
+			$mtg_updte = date("M j, Y  g:i a", $cur_meeting_lastupdate);
```
joshreisner, line 2114: we should use single quotes whenever the string doesn't contain a variable

includes/functions.php
joshreisner: maybe `display: block`? seems strange to render a link as a table

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+		} else {
+			$dow = $db_meeting['day'];
 		}
+
+	    $is_matched = in_array( $mtg_slug, $feed_slugs );
+
+		if (!$is_matched) {
+		    $is_there_change = true;
+			if ( is_numeric( $dow )) {
+				$dayofweek = $days[date($dow)];
```
joshreisner, line 2147: looks like it should just be `$days[$dow]`

includes/functions.php
```diff
@@ -2058,39 +2060,104 @@ function tsml_scan_data_source($data_source_url)
...
+			$mtg_updte = date("M j, Y  g:i a", $cur_meeting_lastupdate);
+			if ( is_numeric( $dow )) {
+				$dayofweek = $days[date($dow)]; 
+			} else {
+				$dayofweek = $dow; 
+			}
+			
+			if ($is_matched) {
+				$message_lines[] = "<tr style='color:gray;'><td>Change</td><td >$mtg_name</td><td>$dayofweek</td><td>$mtg_updte</td></tr>";
+			}
```
joshreisner, line 2123: be great to format this consistently and remove the break, ie `} else {`

joshreisner (changes requested):
flagged a few areas where i think the code could use a little polishing 💅 none of these things are too big but i think they will make it a more attractive PR. things like variable names, unused code, spacing will add up to a nicer coding experience for other developers

includes/functions.php
```diff
@@ -1985,26 +1985,28 @@ function tsml_scan_data_source($data_source_url)
...
+				
+				if (count($meetings_updated) > 0) {
 					// Send Email notifying Admins that this Data Source needs updating
 					$message = "Data Source changes were detected during a scheduled sychronization check with this feed: $data_source_url. Your website meeting list details based on the $data_source_name feed are no longer in sync. <br><br>Please sign-in to your website and refresh the $data_source_name Data Source feed found on the Meetings Import & Settings page.<br><br>";
 					$message .= "data_source_name: $data_source_name <br>";
 					$term = get_term_by('term_id', $data_source_parent_region_id, 'tsml_region');
 					$parent_region = $term->name;
 					$message .= "parent_region: $parent_region <br>";
-					$message .= "change_detect: $data_source_change_detect <br>";
+					//$message .= "change_detect: $data_source_change_detect <br>";
```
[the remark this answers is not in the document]
kiyote33: I did remove it once, but somehow crept back in....

includes/functions.php
```diff
@@ -1985,26 +1985,28 @@ function tsml_scan_data_source($data_source_url)
...
 					$message .= " database count: $data_source_count_meetings <br>";
 					$feedCount = count($meetings);
-					$message .= "import feed cnt: $feedCount<br>";
-					$message .= 'Last Refresh: ' . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
-					if ($meetings_updated) {
-						$message .= "<br><b><u>Detected Difference</b></u><br>";
-						foreach ($meetings_updated as $updated_group) {
-							$message .=  "$updated_group <br>";
-						}
+					$message .= "data source feed cnt: $feedCount<br>";
```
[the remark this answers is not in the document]
kiyote33: I missed that. thx for catching...

includes/functions.php
```diff
@@ -1985,26 +1985,28 @@ function tsml_scan_data_source($data_source_url)
...
+					$message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
+					$message .= "<br><b><u>Detected Difference</b></u><br>";
+					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
+					foreach	($meetings_updated as $key => $value) {
```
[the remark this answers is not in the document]
kiyote33: Ok

includes/functions.php
kiyote33: Thx for the tip. Will use...

includes/functions.php
kiyote33: I had inline-block, but I was attempting to center the link horizontally so tried different methods to do so. Decided not to center the link and missed resetting the display.

includes/functions.php
kiyote33: Yes

includes/functions.php
kiyote33: Will do...

includes/functions.php
kiyote33: I'll try it...

includes/functions.php
joshreisner: it's faster to use `''`. the PHP parser examines the content of every `""` to check for variables inside. it doesn't have to do that for single quotes

includes/functions.php
kiyote33: Your suggesting to not use color to segregate the Update Mode in the Report?

joshreisner: no, i'm suggesting changing 
```
}
else {
```
to 
```
} else {
```

includes/functions.php
kiyote33: Good suggestions. Will do....

includes/functions.php
kiyote33: Good eye. It a left-over from some debugging i did. Will remove...

includes/functions.php
kiyote33: I felt that was pretty weak. I covered only a couple of different scenarios and didn't spend enough time thinking it through. Will put some more thought into it...

includes/functions.php
kiyote33: Another leftover from my debugging code. We can remove...

includes/functions.php
kiyote33: $meetings_updated was legacy code that I hooked into instead of removing. Agree the code is clearer as your suggest.

includes/functions.php
kiyote33: Thx for the tip...

includes/functions.php
```diff
@@ -1994,19 +1994,16 @@ function tsml_scan_data_source($data_source_url)
...
 					$message .= "<br><b><u>Detected Difference</b></u><br>";
 					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
-					foreach	($meetings_updated as $key => $value) {
-						$message .= $value;
-					}
+					$message .= implode('', $meetings_updated);	
 					$message .= "</tbody></table><br>";
-					$import_page_url = 'https://' . $_SERVER['SERVER_NAME'] . '/wp-admin/edit.php?post_type=tsml_meeting&page=import';
-					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: table;font-size: 18px;'>Goto Import & Settings page...</a>";
+					$import_page_url = 'https://' . admin_url() . '/wp-admin/edit.php?post_type=tsml_meeting&page=import';
```
joshreisner, line 2005: `admin_url()` doesn't work this way. you'll get `https://https://mywebsite.org/wp-admin/wp-admin/edit.php…`

try `$import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');` instead

includes/functions.php
```diff
@@ -2080,86 +2074,88 @@ function tsml_import_has_changes($data_source_url, $meetings, $data_source_count
...
 	sort($db_slugs);
-	$db_slugs_cnt = count($db_slugs);
 
-	// list changed and new meetings found in the ds feed
-	foreach ($meetings as $meeting) {
+	// list changed and new meetings found in the data source feed
+	foreach ($feed_meetings as $meeting) {
+	
+		$dayofweek = $dow_number = '';
+		tsml_get_day_of_week_info( $meeting['day'], $dayofweek, $dow_number);
```
joshreisner, line 2085: passing return values by reference is weird - how about

```
list($dayofweek, $dow_number) = tsml_get_day_of_week_info($meeting['day']);
```

includes/functions.php
```diff
@@ -2080,86 +2074,88 @@ function tsml_import_has_changes($data_source_url, $meetings, $data_source_count
...
-		if ($cur_meeting_lastupdate > $data_source_last_refresh) {
-			$is_there_change = true;
-			$mtg_name = $meeting['name'];
-			$mtg_updte = date("M j, Y  g:i a", $cur_meeting_lastupdate);
-			if ( is_numeric( $dow )) {
-				$dayofweek = $days[date($dow)]; 
-			} else {
-				$dayofweek = $dow; 
-			}
+		$feed_slugs[$meeting['slug']] = $meeting_slug;
```
joshreisner, line 2097: this doesn't need to be associative, so i think you just mean
```
$feed_slugs[] = $meeting_slug;
```

includes/functions.php
```diff
@@ -2080,86 +2074,88 @@ function tsml_import_has_changes($data_source_url, $meetings, $data_source_count
...
+
+		// has the meeting been updated since the last refresh?
+		$current_meeting_last_update = strtotime($meeting['updated']);
+		if ($current_meeting_last_update > $data_source_last_refresh) {
+
+			$meeting_name = $meeting['name'];
+			$meeting_update_date = date("M j, Y  g:i a", $current_meeting_last_update);
```
joshreisner, line 2104: `''`

includes/functions.php
```diff
@@ -2080,86 +2074,88 @@ function tsml_import_has_changes($data_source_url, $meetings, $data_source_count
...
+//function:	Returns corresponding day of week string and number for the day input
+if (!function_exists('tsml_get_day_of_week_info')) {
+	function tsml_get_day_of_week_info( $meeting_day_input, &$dayofweek, &$dow_number )
+	{
+		global $tsml_days;
+		
+		// when day is like "Monday" convert to number
+		if ( in_array($meeting_day_input, $tsml_days) ) {
+			$dow_number = date('N', strtotime( $meeting_day_input ) );
+			if ($meeting_day_input == 'Sunday') {$dow_number = '0';}
```
joshreisner, line 2143: sooooo, this would be Sunday = 0, Monday = 2, Tuesday = 3. instead of these lines you could use
```
$dow_number = date('w', strtotime($meeting_day_input));
```
or you should be able to do 
```
$dow_number = array_search($meeting_day_input, $tsml_days);
```

includes/functions.php
```diff
@@ -2080,86 +2074,88 @@ function tsml_import_has_changes($data_source_url, $meetings, $data_source_count
...
+		} else {
+			$dow_number = $meeting_day_input;
+		}
+
+		// only accept valid day of week numbers
+		if (($dow_number >= 0) && ($dow_number <= 6)) {
+		    $dayofweek = $tsml_days[$dow_number];
+		} else {
+		    $dayofweek = 'invalid value';
+		}
```
joshreisner, line 2157: would be cleaner to say
```
if (array_key_exists($dow_number, $tsml_days)) {
```
or the whole thing could be
```
$dayofweek = array_key_exists($dow_number, $tsml_days) ? $tsml_days[$dow_number] : 'invalid value';
```

joshreisner: also if you're going with the `list($dayofweek, $dow_number)` suggestion above, here you'd put a

`return [$dayofweek, $dow_number];`

joshreisner (changes requested):
thanks for making those changes, it's getting closer! 🚂 ⛰️

includes/functions.php
kiyote33: Uhg... I didn't even test it out after making the change. Something came up and I didn't get back to it.  Thanks for catching...

includes/functions.php
kiyote33: Using list was my first attempt which didn't work, so I tried option 2 of the 3 methods I found for returning multiple values from a function call. It seemed to work well. Is that what putting an ampersand in front of the variable does, make it by reference instead of by value? Hmm...

includes/functions.php
kiyote33: I think your right...

includes/functions.php
kiyote33: I was wondering why **$dow_number = date('N', strtotime( $meeting_day_input ) );** returned 1 thru 6 for Monday to Saturday, and 7 for Sunday. I was hoping I could find something better than my hack. I'll give your suggestion a shot tomorrow.  Thanks for the help.

includes/functions.php
kiyote33: Yea, I'm definitely old school here. Ternary operators don't seem clearer to me, but if that's what you all prefer I'll work at it a bit more.
I'll take another run at trying to get ### list($dayofweek, $dow_number) / return [$dayofweek, $dow_number]; working again. Your coding example is pretty much exactly what I tried first and had trouble with. I did testing all afternoon with the 2nd method that I used, so I have some confidence in it.

includes/functions.php
kiyote33: array_search worked like a charm. No more hack...

includes/functions.php
kiyote33: Seems my list problem centered around how I was initializing the variables. Working well now...

includes/functions.php
kiyote33: Tested with parameter in the call. Good to go.

joshreisner (approved):
looks good to me now @kiyote33, thanks for making those changes. hope you don't mind, i went and fixed that call to `tsml_debug` i mentioned in slack, and that automatically PSR-12 formatted the code. you'll want to pull before you make any further changes

brianw-area24 (approved):

PR #797, opened 2022-05-01
Add support for new geocoding gateway

joshreisner:
the only user-facing change is one new option (and one revised option) in the Address Geocoding dropdown on the settings page

Screen Shot 2022-05-01 at 10 49 33 AM

behind the scenes:
* adds support for new gateway
* refactors `tsml_geocode()`
* uses new option called `TSML_ADMIN` that can be set in `wp-config.php`
* creates new Geocoding test page with a lookup tool, as well as gateway registration information and cache displayed

Screen Shot 2022-05-01 at 10 49 04 AM


97eighty (approved):
@joshreisner Great add with the TSML_ADMIN.

tech2serve:
> the only user-facing change is one new option (and one revised option) in the Address Geocoding dropdown on the settings page
> 
> Screen Shot 2022-05-01 at 10 49 33 AM
> 
I'll work to test and review tomorrow.

Regarding drop down options, I believe we can (and should) do away with the old alpha gateway option. It should still work if the user doesn't open the drop down menu, but if they do they should choose between legacy or the new beta option. I surmise very few are using it anyway, and we don't want new potential testers to be confused about which api gateway option to select.

includes/functions.php
```diff
@@ -624,34 +595,97 @@ function tsml_geocode_google($address, $tsml_map_key)
...
+						"keyName" => "appVersion",
+						"value" => TSML_VERSION,
+					], [
+						"keyName" => "wordpressVersion",
+						"value" => get_bloginfo('version'),
+					],
+				]
+			];
+
+			//gateway only supports certain program values
```
tech2serve, line 635: Technically, these are not currently checked on the backend at run-time, and I just slapped a few in to help define them for developers. But, it would be good to have a run-time check here in which case, these are the ones that should be added:

Program types to add:
ACA
CMA
CoDA
CEA-HOW
DA
DAA
GA
HA
OA
PAL
RCA
RD
RR
SAA
SCA
SLAA
VA

And maybe SG?

includes/functions.php
```diff
@@ -624,34 +595,97 @@ function tsml_geocode_google($address, $tsml_map_key)
...
+		$geocode_parameters['apiToken'] = $registration['apiToken'];
 
-	//send request to google
-	if ($tsml_geocoding_method == 'api_gateway') {
-		curl_setopt($tsml_curl_handle, CURLOPT_URL, 'https://api-gateway.apps.itstechnical.net/api/geocode?' . http_build_query($options));
+		curl_setopt($tsml_curl_handle, CURLOPT_URL, 'https://geocode.code4recovery.org/api/v1/geocodes?' . http_build_query($geocode_parameters));
 	} else {
-		curl_setopt($tsml_curl_handle, CURLOPT_URL, 'https://maps.googleapis.com/maps/api/geocode/json?' . http_build_query($options));
+		//use the user's google key if indicated
+		$geocode_parameters['key'] = ($tsml_geocoding_method === 'google_key' && !empty($tsml_google_maps_key))
```
tech2serve, line 678: If the user has a Google mapping key configured, and legacy set, this will default to their mapping key for geocoding, right?

Assuming I'm reading that correctly, I don't believe we want that for legacy as their mapping key might not have geocoding enabled as a service.

tech2serve, line 678: If the user has a Google mapping key configured, and legacy set, this will default to their mapping key for geocoding, right?

Assuming I'm reading that correctly, I don't believe we want that for legacy as their mapping key might not have geocoding enabled as a service.

includes/variables.php
```diff
@@ -105,6 +105,17 @@
 //load the geocoding method
 $tsml_geocoding_method = get_option('tsml_geocoding_method', 'legacy');
 
+//options for geocoding
+$tsml_geocoding_methods = [
+	'legacy' => __('Legacy Method', '12-step-meeting-list'),
+	'api_gateway' =>  __('API Gateway (Alpha, temporary)', '12-step-meeting-list'),
+	'c4r_geocoding' =>  __('API Gateway (Beta)', '12-step-meeting-list'),
```
tech2serve, line 112: I suggest removing `api_gateway` as an option the user can select. Today we only want them testing the new service. We could even automatically shift those users with the alpha test site selected to the new beta?

Second suggestion: Change 112 to `'c4r_geocoding' => __('C4R Geocoding (Beta)', '12-step-meeting-list'),`

tech2serve, line 112: I suggest removing `api_gateway` as an option the user can select. Today we only want them testing the new service. We could even automatically shift those users with the alpha test site selected to the new beta?

Second suggestion: Change 112 to `'c4r_geocoding' => __('C4R Geocoding (Beta)', '12-step-meeting-list'),`

tech2serve (approved):
Looks good overall. I did add a few recommended changes for consideration.

includes/functions.php
joshreisner: Not quite. If a user has entered a Google API Key in the mapping panel then another option appears in the dropdown (after they refresh) called "Use My API Key" with a message:

Screen Shot 2022-05-03 at 7 24 57 AM

joshreisner: If the dropdown menu is set to "Legacy" then they are using our key

includes/functions.php
joshreisner: ok done 👍

includes/variables.php
joshreisner: Screen Shot 2022-05-03 at 7 29 46 AM

ok done 👍

tech2serve (approved):

kiyote33 (commented):
I tried out the new geocoding lookup tool on the new test page, which at first glance appeared to work well. The only unusual thing I noticed was duplicate entries in the Cache for the two locations that I entered in the lookup tool. I didn't expect the lookup tool to add to the Cache. If that is expected behavior, then I wouldn't expect duplicate entries for each lookup. Is this a bug?

joshreisner:
@kiyote33 yes all lookups add to the cache, so that's expected behavior. and i don't think they're duplicate entries, i think they're probably similar. my guess is there's a minor difference in the key that's being looked up. for example, capitalization, or punctuation, or presence of country name

kiyote33 (approved):
Ok. Thanks...

PR #811, opened 2022-05-19
TSML UI distance unit setting to fix #810

### LINKED ISSUE #810: [Bug]: TSML UI: distance unit is mi instead of km while kilometers is set in the settings tab
### Contact Details

adj-webmestre@aa89.org
### Website With Issue

https://aa89.org/wordpress/reunions

### What happened?

the distance displayed in search result should be in km as it is set to kilometers unit in settings General section.
Here in Québec (Canada), mi is not a valid distance unit...
it is only recognized by older people as we use kilometers SI unit since 1970's  ;-)

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

Firefox

_No response_

joshreisner:
distance-unit-km

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -101,24 +101,33 @@ add_shortcode('tsml_types_list', function () {
 //output a react meeting finder widget https://github.com/code4recovery/tsml-ui
 function tsml_ui()
 {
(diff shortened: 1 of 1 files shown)


brianw-area24 (approved):
I don't have location services enabled in browser, so I don't see the miles/km values in tsml_ui, however, I did verify that "distance_units" does get added to the tsml_react_config variable.

Issue #809, opened 2022-05-18
[Bug]: Sync/Import - Regions hierarchy not working


julien-paris:
### Contact Details

ecomm.cer@aamail.org

### Website With Issue

https://alcoholics-anonymous.eu/meetings/

### What happened?

Bug report submitted by CER (Julien) and Berlin (Soren) following extensive testing

As an admin, I am setting up data sources to be pulled into the CER (Continental European Region) website;
I set a parent region;
**Expected behavior**
copies of the source regions should be created or re-used under the selected parent region, regardless of whether they previously exist outside of the parent region or not.

**What happened**

2 issues arise: 

1.a. when I select a parent region wihch **does not** exist in the source website, the regions of the source website become a top-level region in the CER website
1.b. when I select a parent region whch **does** exist in the source website, the regions of the source website become a top-level region in the CER website

2. in both scenarios above, the regions are marked with 0 meetings associated. And in meeting list, the meetings are not associated with the correct region. they are "regionless", until i manually re-save them, in which case they then are associated with their region. 

**Context** 
This has been tested both with "Change detection" on and off. 
The bug also happens with other data sources. 
We have tested both with the region existing and not yet existing
We tested with Berlin as a parent region.
We also tested with Germany as a parent region - which does not exist on the Berlin website.
Both sides using the latest TSML version 3.14.3
The data source: https://www.alcoholics-anonymous-berlin.de/wp-admin/admin-ajax.php?action=meetings&key=39eca0e9954989d7080c24284097fd05

Please don't hesitate to request a zoom/TeamViewer session  in order to dig in. 

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

### What browsers are you seeing the problem on?

Chrome

### Relevant log output/errors

```shell
Example meeting: 
Geocoded:  
Object
approximate: "no"
city: "Berlin"
formatted_address: "Jungstraße 26, 10247 Berlin, Germany"
latitude: 52.5113636
longitude: 13.4659246
status: "cache"
[[Prototype]]: Object
```

kiyote33:
Hi Julien, your bug report looks like it's a bit complicated. As I dive into this I would like to make sure I understand the configuration of your website where the reported issues are occurring.  If I remember correctly, you are using the Hello theme with Elementor. Is all your data imported using our Data Source import utility? Are there multiple data sources in addition to the berlin data source you provided?
[...]

PR #818, opened 2022-06-01
Set parent region on imported data source records.

kiyote33:


kiyote33:
Hi @joshreisner, Sub-domains are inherently supported in the region hierarchy. To achieve what you are suggesting (which I don't believe is what CER really desires), simply import the Berlin data source with no region selected. Once imported, use the Region utility to assign Germany as the parent to the Berlin, Frankfurt (Oder), and Near Berlin regions. That will give you the desired region hierarchy: 
image

joshreisner:
@kiyote33 i think it's safe to say CER desires the regions to be the way they currently are on their site: Germany > Berlin > Kreuzberg.

our goal should be to enable them to achieve that without editing regions post-import

kiyote33:
FYI: I built the current Germany region hierarchy by hand for them so they would have Berlin meetings displaying under Germany while we worked on a resolution to their problem. But I don't think it would be that difficult to add the existing hierarchy to the code, so I will add that in...

kiyote33:
Supports sub_sub_region:
image

joshreisner (approved):
works for me, thanks!

includes/ajax.php
```diff
@@ -339,7 +339,7 @@ function tsml_ajax_geocode()
 
 	$meetings	= get_option('tsml_import_buffer', []);
 	$errors		= [];
-	$limit		= 25;
+	$limit		= 5;
```
tech2serve, line 342: This is a pretty big change, and I might have missed a discussion, but will we see a performance change on large imports?

Apologies if I’m out in left field as I’m working to catch up.

tech2serve (commented):
I’ve not tested the PR in it’s entirety, so I’m not going to provide explicit approval, just wanted to make sure I understood the `$limit` change.

includes/ajax.php
kiyote33: Oversight! Now fixed...

kiyote33:
Fixes issue #809

PR #820, opened 2022-06-04
enabling user-settable location-only column

### LINKED ISSUE #807: [Bug]: Rename columns not working on new UI
### Contact Details

clintjohnson@nmbgeek.com

### Website With Issue

https://aamyrtlebeach.org

### What happened?

Enabled the new UI and wanted to simplify the column names (remove / in Location and Address columns. Added this to the functions.php with no luck:

```
$tsml_columns = array(
	'time' => 'Time',
	'name' => 'Name',
	'types' => 'Types',
	'location' => 'Location',
	'address' => 'Address',		
	'distance' => 'Distance',
	'region' => 'Region'
);
```

The JS hack below does work but because wp_footer still ends up above the TSML code I have to set a timeout to execute the script so that it doesn't happen before TSML is loaded.  I tried onload and a few other methods and this was the only way I could get it to work.  Of course this causes the column title to change after the user sees the original naming.

```
function tsml_rename_columns() {
  $page = get_page_by_path("meetings");
  if ($page) {
    ?>
        <script>
		setTimeout(function () {
			document.querySelector("th.location_group").innerHTML = "Location";
			document.querySelector("th.address").innerHTML = "Address";
		}, 1000);
        </script>
    <?php
  }
}
add_action('wp_footer', 'tsml_rename_columns');
```

Thank you all for your great work.

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

Firefox, Chrome, Microsoft Edge

```shell
I really just want to rename the columns so that they do not have slashes.
```

### LINKED ISSUE #819: add "location" as a user-settable column
### Contact Details

_No response_

### Requested Feature/Enhancment

enable people to have a location-only (as opposed to location/group) column in their meeting lists

this would be via adding PHP code to their theme and not be default behavior

this will unblock #807 

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

joshreisner:
this PR will close #819 and also close #807 

it adds the ability for users to make their meeting list display location only via adding 

```
$tsml_columns = [
	'time' => 'Time',
	'distance' => 'Distance',
	'name' => 'Meeting',
	'location' => 'Location',
	'address' => 'Address',
	'region' => 'Region',
	'district' => 'District',
	'types' => 'Types',
];
```

It also propagates these preferences to TSML UI, if that is selected

Bonus: it syncs the type descriptions as well, for sites with non-aa programs selected


kiyote33 (approved):
Works good, both TSML & TSML UI...

PR #824, opened 2022-06-09
Updates for version 3.14.4

brianw-area24:
Prep to release version 3.14.4


brianw-area24:
I've updated the tested up to value as per the suggestion

PR #836, opened 2022-06-20
fixing filtering for the location/group column

joshreisner:
see https://github.com/code4recovery/12-step-meeting-list/discussions/835


97eighty (approved):

kiyote33 (commented):
Looks good. Location/Group column data displaying as expected in my test site.

PR #859, opened 2022-07-07
Fix build of directions URL for Google Maps 

tim-rohrer:
See Discussion #857 .

Google Maps has changed the way URLs are built for getting directions. This PR updates TSML to the current Google guidance.

This was tested with Chrome on a laptop. It still needs to be tested on a site using a mobile device.


brianw-area24 (approved):
Changes look good to me

Issue #873, opened 2022-07-17
[Bug]: Import Of Non Unique Slugs In Data Source Cannot Be Matched in Change Detection


kiyote33:
### Contact Details

waynep@code4recovery.org

### What happened?

Process to repeat this bug:

1.  Import a meeting called "Big Book Study" with a big-book-study slug from Site A (https://district1.multisite.dev).
2.  Import a meeting called "Big Book Study" with a big-book-study slug from Site B (https://district2.multisite.dev).
3.  The 2nd import process changes the meeting slug to big-book-study-2 to maintain key uniqueness in the database.
4.  When Data Source Change Detection is enabled, the big-book-study-2 record is reported as "Removed" in the Change Detection Report sent to the site admin, as there is no match for that key in the data source feed. Here is the report from a test site:
![image](https://user-images.githubusercontent.com/57125427/179421645-e253358c-bc41-4c78-8a3f-be6c5a953fd4.png)

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

### What browsers are you seeing the problem on?

Microsoft Edge

Issue #872, opened 2022-07-17
[Feature Request]: Add Hypertext Reference to Change Detection Report Meeting Name


kiyote33:
### Contact Details

waynep@code4recovery.org

### Requested Feature/Enhancment

**Please add this cool feature!** Convert the Meeting Name text field to include a hypertext reference so explicit details of a reported meeting change can be quickly accessed. i.e.: $meeting_name = "<a href='get_permalink($meeting->ID)'> $meeting_title </a>"

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

PR #874, opened 2022-07-17
#872 Add href link to meeting name.

kiyote33:
#873 Fix to match modified slugs to import feed slugs.


97eighty:
This change tests out for me. Only issue is it not being a clean fix. Will work for -1 through -4 but what if I have several more? Perhaps a loop?

kiyote33:
Hi Mark,

 

I’ve put the fix in a 10 loop which makes for a little cleaner result. Could you test and approve if OK?

 

Thanks…

Wayne

97eighty (approved):

PR #933, opened 2022-09-19
Export the data_source

douggreen:
# Description

It would be nice to keep track of the source of meetings. For example, if a district website is the source of truth for the meeting, if  this data is imported by both the are, and if GSO imports the data from the area, then GSO automatically assumes that the source is the area and not the district. We already have a field for the data_source, that can be used and is in the export, but it's not set on the original source.

This change will set the source on the original site (in the example above, on the district site) so that when other sites export and import the meeting, that the data source is known.

Once this change is merged, the next thing to do is to coordinate with GSO and the Meeting Guide app to use the data.


douggreen:
Ok, my mistake, When I wrote this I was thinking that the import came from the CSV, but it comes from the JSON (and I knew this). We need to change the JSON. i'll re-work this.

PR #966, opened 2022-11-01
964 update registration of tslm_meeting post type with with_front filter

mdavoli:
Added filter to allow altering the `with_front` configuration of the `tsml_meeting` post type. Also updated documentation with instructions on how to apply the filter.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -300,7 +300,7 @@ function tsml_custom_post_types()
 			'show_ui' => true,
 			'has_archive' => $is_public,
 			'menu_icon' => 'dashicons-groups',
-			'rewrite' => ['slug' => $tsml_slug],
+			'rewrite' => ['slug' => $tsml_slug, 'with_front' => apply_filters('tslm_meeting_with_front', true)],
 		]
 	);
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -300,7 +300,7 @@ function tsml_custom_post_types()
 			'show_ui' => true,
 			'has_archive' => $is_public,
 			'menu_icon' => 'dashicons-groups',
-			'rewrite' => ['slug' => $tsml_slug, 'with_front' => apply_filters('tslm_meeting_with_front', true)],
+			'rewrite' => ['slug' => $tsml_slug, 'with_front' => apply_filters('tsml_meeting_with_front', true)],
 		]
 	);
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -300,7 +300,7 @@ function tsml_custom_post_types()
 			'show_ui' => true,
 			'has_archive' => $is_public,
 			'menu_icon' => 'dashicons-groups',
-            'rewrite' => ['slug' => $tsml_slug, 'with_front' => apply_filters('tsml_meeting_with_front', true)],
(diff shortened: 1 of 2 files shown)


97eighty (approved):
Should the permastruct be prepended with the front base. (example: if your permalink structure is /blog/, then your links will be: false->/meetings/, true->/blog/meetings/). Defaults to true.

mdavoli:
@97eighty  that's correct. I tested locally and confirmed that behavior

includes/functions.php
```diff
@@ -300,7 +300,7 @@ function tsml_custom_post_types()
 			'show_ui' => true,
 			'has_archive' => $is_public,
 			'menu_icon' => 'dashicons-groups',
-			'rewrite' => ['slug' => $tsml_slug],
+			'rewrite' => ['slug' => $tsml_slug, 'with_front' => apply_filters('tslm_meeting_with_front', true)],
```
joshreisner, line 303: we use PSR-12, looks like this might need some more indentation here. could you format this with PHP Intelephense?

readme.txt
```diff
@@ -249,6 +249,11 @@ You may set it to false to hide the public meeting finder altogether.
 
 To apply these changes, you must go to Settings > Permalinks and click "Save Changes"
 
+= Can I change the with_front configuration of the tsml meeting post type
```
joshreisner, line 252: Could you change this to be the problem the user would be trying to solve by changing the configuration?

readme.txt
```diff
@@ -249,6 +249,11 @@ You may set it to false to hide the public meeting finder altogether.
...
+Yes, you can use the following filter to change the with_front configuration from true to false
+
+	add_filter( 'tslm_meeting_with_front', '__return_false');
```
joshreisner, line 255: could you change the filter name to start with our prefix `tsml_`

joshreisner (changes requested):

readme.txt
mdavoli: Done, sorry about the typo

readme.txt
mdavoli: I've updated the copy, let me know if that's ok.

includes/functions.php
mdavoli: I think this is resolved but not 100% sure, please confirm.

joshreisner (approved):

joshreisner:
hey @mdavoli we'll go ahead and merge this for the next release. could you describe the problem though? i'm seeing the cure but not the symptom.

is it that you have a site at `mywebsite.org`, and your blog is at `mywebsite.org/blog`, and this helps you make your meetings page be at `mywebsite.org/meetings` instead of `mywebsite.org/blog/meetings`?

mdavoli:
@joshreisner  thanks for merging, appreciate it. You are exactly right, in your description. The primary site is a `mywebsite.org` but the blog lives in a `mywebsite.org/blog` like location (due to permalink configuration)  but they wanted to keep the URL for the meetings to `mywebsite.org/meetings` which wasn't possible before this fix. Allowing us to override the `with_front` setting solves the problem.

You can recreate this problem by going to the permalinks setting of `custom structure` and put in something like this: `/blog/%topic%/%postname%/`

Via code pages and other custom post types were configured `with_front` to be false so they all were at site root of `mywebsite.org`

PR #970, opened 2022-11-10
update directions urls

joshreisner:
fixes directions URLs on iOS while tweaking address on other platforms using Google's latest guidance


kiyote33 (approved):
Directions now work as expected on both my windows desktop and android phone with this fix. Good work Josh...

PR #971, opened 2022-11-11
3.14.7 release prep

joshreisner:


kiyote33 (approved):
Yep. Two prep files look good and the directions fix merged correctly. Tested on Windows desktop and Android phone. Good to go...

PR #1084, opened 2023-03-01
add custom descriptions function

### LINKED ISSUE #1083: tsml_custom_descriptions
### Contact Details

_No response_

### Requested Feature/Enhancement

we should have a function like `tsml_custom_types` and `tsml_custom_flags` so that admins can customize type descriptions when necessary

joshreisner:
the plugin doesn't have a helper function to change the type descriptions, like it does for types, addresses, and flags

closes #1083


kiyote33 (approved):
Worked like a charm when I dropped example code in a child theme functions.php file. Very useful override function.

Issue #1073, opened 2023-02-11
Feeds import puts 12x12 meetings in notes


joshreisner:
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1071

<div type='discussions-op-text'>

<sup>Originally posted by **anchovie91471** February  9, 2023</sup>
### Contact Details

anthony@bustedspring.com

### What happened?

First, thanks for the amazing plugin and all the hard work you have put in for the recovery community!! I used the plugin on several sites for districts & our area. It's a huge help!

I identified a bug today in the latest version:

When importing meetings from a JSON feed. The 12x12 meeting type gets imported into the Meeting Notes field instead of the meeting types array.

This appears to be due to a failing if statement on line 1521 of includes/functions.php.
`if (array_key_exists($upper_type, $upper_types))`

The key for all 12x12 meetings in the variables.php file is set to "12x12" (lowercase x) instead of "12X12" so it fails this test and gets put in the notes field.

I sent an email a week or two ago to info@code4recovery.org asking for info regarding how best to do pull requests but didn't hear back. I'd be happy to do the PR and fix if that's helpful.

### Relevant log output/errors

_No response_</div>

97eighty:
Is there a source JSON feed to test with?

joshreisner:
sure here's one: https://sheets.code4recovery.org/storage/12KfpoNcLP4YqRsiN_jakbaxkF-ZQpcqTncISQ5Fr0e4.json

if i import it to our test site: https://code4recovery.org/meetings/odaat-2

PR #1087, opened 2023-03-14
Feeds import puts 12x12 meetings in notes

anchovie91471:
When importing meetings from a remote feed, the "12x12" meeting type is inserted into the meeting notes field, instead of importing into the "types" meta key.

This PR fixes this by changing the condition that checks the incoming feed types against the known types on the target website.

**Original code**
Previously, the check (inside the tsml_import_buffer_set() function, line 1535) converted types from the origin feed to uppercase and checked against the keys of known types on the target site, which is not uppercase for "12x12" meetings, according to the TSML spec.

**The update fixes this by:**
1. Comparing the same uppercase types from the origin feed against an array of uppercase keys of known types, converting "12x12" to "12X12", thereby passing the check, and...
2. Setting the type key that is eventually saved to the incoming type (12x12) instead of the uppercased type (12X12), preserving the correct key, according to the TSML Spec.

closes code4recovery/12-step-meeting-list#1073

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1532,8 +1532,8 @@ function tsml_import_buffer_set($meetings, $data_source_url = null, $data_source
 		$meetings[$i]['types'] = $unused_types = [];
 		foreach ($types as $type) {
 			$upper_type = trim(strtoupper($type));
-			if (array_key_exists($upper_type, $upper_types)) {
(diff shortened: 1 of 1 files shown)


joshreisner (approved):
works great!

PR #1096, opened 2023-03-20
Fix [tsml_types_list] & [tsml_regions_list] shortcodes.

### LINKED ISSUE #1091: URL's in shortcodes [tsml_types_list] & [tsml_regions_list] are incorrect when using TSML UI as frontend
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1086

<sup>Originally posted by **anchovie91471** March 13, 2023</sup>
I just noticed this when using the shortcodes on a fresh WP 6.1.1 install with only TSML installed, using the 2023 theme. (I tried 2021 as well)

The links are fine for Legacy UI, but with TSML UI (react), the parameters look like they should be different.

1. **[tsml_types_list]:** The link for "Big Book" is [https://northmsaa.org/testsite/meetings/?tsml-day=any&tsml-type=B](https://northmsaa.org/testsite/meetings/?tsml-day=any&tsml-type=B) but should be (maybe) [https://northmsaa.org/testsite/meetings/?tsml-type=big-book](https://northmsaa.org/testsite/meetings/?tsml-type=big-book). Looks like TSML UI uses a slug instead of the type key.
2. **[tsml_regions_list]:** The link for a test region I have (tag_id=8] is [https://northmsaa.org/testsite/meetings/?tsml-day=any&tsml-region=8](https://northmsaa.org/testsite/meetings/?tsml-day=any&tsml-region=8) but should be [https://northmsaa.org/testsite/meetings/?region=fulton-ms](https://northmsaa.org/testsite/meetings/?region=fulton-ms). Again, looks like TSML UI expects a slug instead of the term ID.</div>

### LINKED ISSUE #1094: Notice shown from upcoming meetings widget when in person meeting has no types
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1093

<sup>Originally posted by **district6wc** March 16, 2023</sup>
I was working on my site tonight and when I updated my homepage this appeared on my sidebar:

Notice: Undefined index: types in /home/izgr1zclaxch/public_html/wp-content/plugins/12-step-meeting-list/includes/shortcodes.php on line 30
Notice: Undefined index: types in /home/izgr1zclaxch/public_html/wp-content/plugins/12-step-meeting-list/includes/shortcodes.php on line 36

This is my site https://alanonalateen6nc.org/

How do I correct this?

Rose L
District 6</div>

anchovie91471:
**Notice: Undefined index: types shown from upcoming meetings widget when in person meeting has no types.**

This PR fixes this issue by adding array_key_exists('types', $meeting) conditionals before assigning the $classes & $meeting_types variables.

closes #1094 

**[tsml_types_list] & [tsml_regions_list] shortcodes generate URL's that don't work with TSML UI**

I took the liberty to bring the filters on the Legacy UI into parity with TSML UI to fix the issue, instead of keeping 2 differing URL patterns. 

Instead of having:
`&tsml-type=B` for Legacy UI and `&type=big-book` for the React interface 

AND

`tsml-region=[region numeric id]` for Legacy and `region=region-name` for React 

I refactored code so the Legacy UI filters use the same patterns as TSML UI.

This required changes to public.js (and thereby public.min.js) to remove "tsml-" from region & type in query strings. I also added had to add decodeURIComponent() function to the type to deal with the possibility that an encoded ampersand (%26) will appear for "12 Steps & 12 Traditions" type.

There were also some changes to the meetings archive template that may be significant.

I also added PHPDoc comments where I created new functions or properties, added them to the filter_meetings class and formatted code according to Wordpress coding standards anywhere I touched so you'll see some differences there.

Lastly, I'm not sure how my branch got behind, but I had to do a final commit to add the list() functions back into the tsml_import_changes() function in functions.php so that wasn't part of this PR. I had no intention/need of that changing.

closes #1091


includes/functions.php
```diff
@@ -2130,7 +2130,7 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 
 	// list changed and new meetings found in the data source feed
 	foreach ($feed_meetings as $meeting) {
-
+
```
anchovie91471, line 2133: This was re-added because my repo got behind.

joshreisner:
i don't want to speak unilaterally for the group, but i don't think it's advisable to alter the existing URL structure, since untold hundreds of websites (and thousands of users) may depend on the URLs being how they are.

these `tsml-` prefixes were originally added to prevent conflict with [wp reserved terms](https://codex.wordpress.org/Reserved_Terms), such as `types`. this i believe can cause problems if the website URL scheme is in the default query string mode

PR #1097, opened 2023-03-21
Fixing undefined index error on upcoming meetings shortcode

### LINKED ISSUE #1094: Notice shown from upcoming meetings widget when in person meeting has no types

anchovie91471:
When in-person meeting has no types, php throws Notice: Undefined index: types

Added array_key_exists check inside tsml_next_meetings() function.

closes #1094

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -27,18 +27,16 @@ function tsml_next_meetings($arguments)
 	$meetings = array_slice($meetings, 0, $arguments['count']);
 	$rows = '';
 	foreach ($meetings as $meeting) {
-		if (array_key_exists('types', $meeting))
-			$classes = tsml_to_css_classes($meeting['types']);
-
+		$meeting_types = $classes = '';
+		if ( ! empty( $meeting['types'] ) ) {
+			$classes       = tsml_to_css_classes( $meeting['types'] );
+			$meeting_types = ' <small><span class="meeting_types">' . tsml_format_types( $meeting['types'] ) . '</span></small>';
+		}
+		
 		if (!empty($meeting['notes'])) {
 			$classes .= ' notes';
 		}
-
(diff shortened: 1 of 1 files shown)


includes/shortcodes.php
```diff
@@ -27,13 +27,14 @@ function tsml_next_meetings($arguments)
 	$meetings = array_slice($meetings, 0, $arguments['count']);
 	$rows = '';
 	foreach ($meetings as $meeting) {
-		$classes = tsml_to_css_classes($meeting['types']);
+		if (array_key_exists('types', $meeting))
+			$classes = tsml_to_css_classes($meeting['types']);
```
joshreisner, line 31: looks good! two small suggestions:

when dealing with undefined key notices, we typically use `if (!empty(` as below because it won't throw a notice and will also tell us if there's a value there.

also i wonder if this could be combined with the two other if statements here like this:

```
$meeting_types = '';
if (!empty($meeting['types'])) {
  $classes = tsml_to_css_classes($meeting['types']);
  $meeting_types = tsml_format_types($meeting['types']);
}
```

then we could get rid of lines 37-40

anchovie91471: OK, Just committed that.

joshreisner (approved):
works great, thanks! changes are approved, so feel free to merge whenever

PR #1102, opened 2023-03-24
Fix URLs in [tsml_types_list] & [tsml_regions_list] shortcodes

### LINKED ISSUE #1091: URL's in shortcodes [tsml_types_list] & [tsml_regions_list] are incorrect when using TSML UI as frontend

anchovie91471:
Conditionally changes the URL's on both shortcodes, so both TSML_UI and LEGACY_UI work with them.

I used the tsml_meetings_url() function to generate the URL's instead of setting manually with $base.

I didn't save the entire file as PSR-12, just the 2 functions I edited. I assume all those tabs are from my last commit? 

closes #1091

### DIFF
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -85,16 +85,21 @@ function tsml_next_meetings($arguments)
 add_shortcode('tsml_next_meetings', 'tsml_next_meetings');
 
 //output a list of types with links for AA-DC
-
 add_shortcode('tsml_types_list', function () {
-	global $tsml_types_in_use, $tsml_programs, $tsml_program;
-	$types = [];
-	$base = get_post_type_archive_link('tsml_meeting') . '?tsml-day=any&tsml-type=';
-	foreach ($tsml_types_in_use as $type) {
-		$types[$tsml_programs[$tsml_program]['types'][$type]] = '<li><a href="' . $base . $type . '">' . $tsml_programs[$tsml_program]['types'][$type] . '</a></li>';
-	}
-	ksort($types);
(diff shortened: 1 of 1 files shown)


joshreisner (approved):
🥇

Issue #1075, opened 2023-02-12
Import Data Sources Redesign with Change Detection


kiyote33:
### Contact Details

@kiyote33

### Requested Feature/Enhancement

Improve the Import Data Sources feature, allowing local admins to apply only the changes detected between an import feed and the local database. The current method which does a complete removal and re-entry of all feed records would no longer be necessary when refreshing. This proposed modification will not affect the existing methods of removing a data source's entire set of records from the local database. The option to enable Change Detection email reports will remain the same by including at least 1 Change Notification email recipient address on the Settings page. 

Dev Notes:
1. Change detection for Data Sources Import feeds needs to be more robust then jus relying on the "updated" field, as it may not be available on all imports. Instead, we will use a database matching record comparison to determine import update candidate records which are new, changed, or removed from the import. This type of differential updating would avoid the necessity of requiring certain fields (i.e. updated) to be present and accurate in the import feed records.
2. Each Data Sources Import feed is registered in the database with its feed URL ($data_source_url) appended to the import record in the database data_source field. This supports the updating and removal of those records in future operations. 
3. The Data Source Name field currently allows admins to name their feed import whatever they want. 
4. The current Import CSV feature will remain unchanged.
5. A debug feature will be included to give admins and developers the ability to see what changes have been detected and applied by the refresh operation. [$tsml_debug = true;]
6. The only perceivable change to the user interface for this feature will be minimal changes to the Import Data Sources text content describing high-level use of the feature.

joshreisner:
i’m all for making the update more efficient. why should we keep the old way as an option? would there be a different outcome between the two methods?

foundartfactory:
FWIW, I am all for this requested feature! I could maybe see wanting to be able to one-off refresh the database w/the entire feed contents -- but if that went away, I would miss it less than I would appreciate the new incremental feature.

joshreisner:
again, im not clear what the benefit of having two methods would be. wouldnt the end result be exactly the same?

having two methods implies that we think the new way isnt 100% reliable, which should not be the case if we are comfortable shipping it

kiyote33:
@joshreisner - I agree. The refresh process will always be there if the feed is dropped and then recreated. We wouldn't need to change that code (other than to change what code executes when the button is pressed). Much cleaner to modify the button text from "Refresh" to "Update" and have the button execute the new code to update the detected changes only.

[...]

PR #1103, opened 2023-03-24
#1075 Update Data Source Changes Only

kiyote33:
Ready for review.


joshreisner:
cool! it would be great to test this out on a demo site, maybe at code4recovery.org? something we could talk about tomorrow

does this do geocoding when addresses change?

kiyote33:
@joshreisner - I was thinking about how best to demo. I will do a preliminary run on code4recovery to see if I can quickly show the effect of all the changes. And yes, it does do geocoding exactly like before, but not just when the address changes. I re-thought that piece and decided the cost/benefit wasn't there for the amount of new code I would be introducing.

includes/admin_import.php
```diff
@@ -167,10 +167,40 @@ function tsml_import_page()
...
+				$feed_updates = $db_ids_to_delete = $message_lines = [];
+				$message = "<br><h2>Local Database Updates Being Applied From the $data_source_name Feed</h2><br>";
+				$data_source_last_import = null;
+
+				//if already set, we want to remove only db records which have been changed or removed from the feed
 				if (array_key_exists($data_source_url, $tsml_data_sources)) {
-					tsml_delete(tsml_get_data_source_ids($data_source_url));
+
+					$tsml_data_sources = get_option('tsml_data_sources', array());
+					$data_source_last_import = (int) $tsml_data_sources[$data_source_url]['last_import'];
```
joshreisner, line 181: i think `intval` is a more standard way to cast this as an int

includes/admin_import.php
```diff
@@ -167,10 +167,40 @@ function tsml_import_page()
 			]);
 			if (is_array($response) && !empty($response['body']) && ($body = json_decode($response['body'], true))) {
 
-				//if already set, hard refresh
+				/*When a data source already exists we want to apply changes detected to the local db */  
+
+				//initialize variables 
+				$feed_updates = $db_ids_to_delete = $message_lines = [];
+				$message = "<br><h2>Local Database Updates Being Applied From the $data_source_name Feed</h2><br>";
```
joshreisner, line 174: be great to localize this string in a `__()`

includes/admin_import.php
```diff
@@ -167,10 +167,40 @@ function tsml_import_page()
...
+					foreach ($db_ids_to_delete as $id) {
+						tsml_delete_by_id($id);
+					} 
 					tsml_delete_orphans();
+					//rebuild cache
+					tsml_cache_rebuild();
+
+					if (count($feed_updates) === 0) {
+						$last_update_date = date('M j, Y  g:i a', $data_source_last_import);
+						exit('<br><h2>No Updates Available For Your ' . $data_source_name . ' Feed</h2>Your local database meeting list records derived from the ' . $data_source_name . ' feed are already in sync!<br>Press the browser back button to return to the previous screen.<br>' );
```
joshreisner, line 196: would be great to localize. also it's not great UI to simply die on a white screen, could we redirect back with a notice?

includes/admin_import.php
```diff
@@ -167,10 +167,40 @@ function tsml_import_page()
...
+					}
+					else {
+						$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
+						$message .= implode('', $message_lines);
+						$message .= "</tbody></table>";
+						print_r($message);
```
joshreisner, line 202: is this `print_r` for debugging?

includes/admin_import.php
```diff
@@ -183,14 +213,20 @@ function tsml_import_page()
 					'type' => 'JSON',
 				];
 
-				//import feed
-				tsml_import_buffer_set($body, $data_source_url, $data_source_parent_region_id);
+				if (count($feed_updates) === 0) {
+					//import feed
+					tsml_import_buffer_set($body, $data_source_url, $data_source_parent_region_id);
+				}
+				else {
```
joshreisner, line 220: would be great to format this with intelephense

includes/admin_import.php
```diff
@@ -284,8 +320,8 @@ function tsml_import_page()
...
 				Data sources listed below will pull meeting information into this website. A configurable schedule allows for each enabled data source to be scanned at least once per day looking 
 				for updates to the listing. Change Notification email addresses are sent an email when action is required to re-sync a data source with its meeting list information. 
-				Please note: records that you intend to maintain on your website should always be imported using the Import CSV feature below. <b>Data Source records will be overwritten when the 
-				parent data source is refreshed.</b> More information is available at the <a href="%s" target="_blank">Meeting Guide API Specification</a>.', '12-step-meeting-list'), 'https://github.com/code4recovery/spec') ?>
+				Please note: records that you intend to maintain on your website should always be imported using the Import CSV feature below. <b>Data Source records will be overwritten when an update from the 
+				parent data source is applied.</b> More information is available at the <a href="%s" target="_blank">Meeting Guide API Specification</a>.', '12-step-meeting-list'), 'https://github.com/code4recovery/spec') ?>
```
joshreisner, line 324: not sure i understand the purpose of rebranding "refresh" to "update" here and below. when a user clicks the button there might not be any updates, they are merely syncing their data with the feed - nothing about this has changed as part of this PR afaik

includes/functions.php
```diff
@@ -227,6 +227,14 @@ function tsml_custom_addresses($custom_overrides)
 	$tsml_google_overrides = array_merge($tsml_google_overrides, $custom_overrides);
 }
 
+//fuction:	define custom type descriptions
+//used:		theme's functions.php
+function tsml_custom_descriptions($descriptions)
```
joshreisner, line 232: thing you might need to merge the branch with main

includes/functions.php
```diff
@@ -383,6 +391,22 @@ function tsml_delete($post_ids)
 	tsml_cache_rebuild();
 }
 
+//function:	remove a meeting
+//used:	admin-import.php
+function tsml_delete_by_id($meeting_id)
```
joshreisner, line 396: `tsml_delete` should be able to handle this

i think `tsml_delete_by_id(123)` is the same as `tsml_delete([123])`

includes/functions.php
```diff
@@ -2022,21 +2046,21 @@ function tsml_scan_data_source($data_source_url)
 
 				if (count($meetings_updated) > 0) {
 					// Send Email notifying Admins that this Data Source needs updating
-					$message = "Data Source changes were detected during a scheduled sychronization check with this feed: $data_source_url. Your website meeting list details based on the $data_source_name feed are no longer in sync. <br><br>Please sign-in to your website and refresh the $data_source_name Data Source feed found on the Meetings Import & Settings page.<br><br>";
+					$message = "Data Source changes were detected during a scheduled sychronization check with this feed: $data_source_url. Your website meeting list details based on the $data_source_name feed are no longer in sync. <br><br>Please sign-in to your website and update the $data_source_name Data Source feed found on the Meetings Import & Export page.<br><br>";
```
joshreisner, line 2049: is this text current? users don't need to sign in and update now right - the update has already been applied?

includes/functions.php
```diff
@@ -2022,21 +2046,21 @@ function tsml_scan_data_source($data_source_url)
...
-					$message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
+					$message .= "Last Update: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
 					$message .= "<br><b><u>Detected Difference</b></u><br>";
 					$message .= "<table border='1' style='width:600px;'><tbody><tr><th>Update Mode</th><th>Meeting Name</th><th>Day of Week</th><th>Last Updated</th></tr>";
 					$message .= implode('', $meetings_updated);
 					$message .= "</tbody></table><br>";
 					$import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');
-					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Settings page</a>";
+					$message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Export page</a>";
```
joshreisner, line 2063: nice 😎

includes/functions.php
```diff
@@ -2141,7 +2165,8 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 		$feed_slugs[] = $meeting_slug;
 
 		// has the meeting been updated since the last refresh?
-		$current_meeting_last_update = strtotime($meeting['updated']);
+		$current_meeting_localised_last_update = tsml_date_localised(get_option('date_format') . ' ' . get_option('time_format'), strtotime($meeting['updated']));
+		$current_meeting_last_update = strtotime($current_meeting_localised_last_update);
```
joshreisner, line 2169: if we are using this to compare to a UNIX date below i don't think it should be localized. i think we can just do `time()` which should return the current unix timestamp

includes/functions.php
```diff
@@ -2269,7 +2294,7 @@ function tsml_date_localised($format, $timestamp = null)
 	// From https://mediarealm.com.au/
 
 	$tz_string = get_option('timezone_string');
-	$tz_offset = get_option('gmt_offset', 0);
+	$tz_offset = get_option('gmt_offset', -6);
```
joshreisner, line 2297: my eye was drawn to the `-6` here but im not sure i undertand the purpose of this `tsml_strtotime` function as a whole, are we sure we need it?

includes/functions.php
```diff
@@ -2294,4 +2319,111 @@ function tsml_date_localised($format, $timestamp = null)
...
+			$is_matched = in_array($meeting_slug, $db_slugs);
+		}
+
+		// add slug to feed array to help determine current db removals later on...
+		$feed_slugs[] = $meeting_slug;
+
+		// has the meeting been updated since the last update?
+		$current_meeting_localised_last_update = tsml_date_localised(get_option('date_format') . ' ' . get_option('time_format'), strtotime($meeting['updated']));
+		$current_meeting_last_update = strtotime($current_meeting_localised_last_update);
+		if ($current_meeting_last_update > $data_source_last_update) {
```
joshreisner, line 2377: again i think `time()` could eliminate the need for getting any options

kiyote33:
What would that code look like?

includes/functions.php
```diff
@@ -2294,4 +2319,111 @@ function tsml_date_localised($format, $timestamp = null)
...
+	$ds_ids = tsml_get_data_source_ids($data_source_url);
+	sort($ds_ids);
+
+	/* filter out all but the data source meetings  */
+	foreach ($all_db_meetings as $db_meeting) {
+		$db_id = $db_meeting['id'];
+		if (in_array($db_id, $ds_ids)) {
+			array_push($db_meetings, $db_meeting);
+		}
+	}
```
joshreisner, line 2349: i think using `array_filter` might be a nice improvement here, instead of making a new array with `foreach`

includes/functions.php
```diff
@@ -2294,4 +2319,111 @@ function tsml_date_localised($format, $timestamp = null)
...
+	foreach ($feed_meetings as $meeting) {
+
+		list($day_of_week, $dow_number) = tsml_get_day_of_week_info($meeting['day'], $week_days);
+		$meeting_slug =  $meeting['slug'];
+
+		// match feed/database on unique slug
+		$is_matched = in_array($meeting_slug, $db_slugs);
+
+		if (!$is_matched) {
+			// numeric slugs may need some reformatting
```
joshreisner, line 2364: not sure i understand why numeric slugs are different?

includes/admin_import.php
kiyote33: Legacy code, but I'm willing to learn. What's the advantage of intval over int?

includes/admin_import.php
kiyote33: Absolutely. Will do.

includes/functions.php
```diff
@@ -2294,4 +2319,111 @@ function tsml_date_localised($format, $timestamp = null)
...
+
+		// add slug to feed array to help determine current db removals later on...
+		$feed_slugs[] = $meeting_slug;
+
+		// has the meeting been updated since the last update?
+		$current_meeting_localised_last_update = tsml_date_localised(get_option('date_format') . ' ' . get_option('time_format'), strtotime($meeting['updated']));
+		$current_meeting_last_update = strtotime($current_meeting_localised_last_update);
+		if ($current_meeting_last_update > $data_source_last_update) {
+			$meeting_name = $meeting['name']; 
+			$meeting_update_date = date('M j, Y  g:i a', $current_meeting_last_update);
```
joshreisner, line 2379: since this date is for output to the user, this actually seems like a place where the site date format option would be appropriate

includes/admin_import.php
kiyote33: Again, willing to learn. I didn't see that option with the exit function. Can you advise what that code looks like?

includes/admin_import.php
joshreisner: as discussed on zoom (but for anyone else looking) i think `intval` is more resilient to unexpected values

includes/admin_import.php
joshreisner: as discussed, i'll handle formatting in release prep

includes/functions.php
joshreisner: i like to use the [github desktop app](https://desktop.github.com/) which makes this super easy. but from the command line i think you can just do `git merge main` and then push the change with `git push origin` and this change should go away

includes/functions.php
joshreisner: nevermind, i was confused ;)

includes/admin_import.php
kiyote33: No. Using to display feedback.

includes/admin_import.php
kiyote33: Check...

includes/admin_import.php
kiyote33: tomato, tomato...

includes/functions.php
kiyote33: I struggled with passing it the array of ids, so I did it this way to move on. I'll revisit and get it to work...

includes/functions.php
kiyote33: ??? Don't understand this question. This is the Change Detection code for the email sent to the registered change notification addresses. No updates have been applied to the local database yet!

includes/functions.php
kiyote33: Nope, not comparing to the current timestamp. We are comparing the last update field (which is UTC based) of the current record to the database timestamp of the last refresh which is stored as a localized datetime field.

includes/functions.php
kiyote33: Good catch. I was trying something and it didn't pan out. I need to reset to 0 or better yet, if it's not being used anywhere (which I think is the case) then just get rid of it.

includes/functions.php
kiyote33: Not sure how that would work. This might be a teachable moment. Could I get an example of how that would look?

includes/functions.php
kiyote33: Can't remember why, but that came up in my initial round of testing on Change Detection using a data source that used database generated numeric ids for slugs. This code solved the problem.

includes/functions.php
kiyote33: Good Idea...

includes/functions.php
kiyote33: ???

includes/admin_import.php
kiyote33: (int) Typecast / intval()
Typecasting in PHP is a way to force a value into a format such as an integer without condition or recourse. This means that typecasting with (int) is a form of sanitization but not validation and thus does not provide a message when the value passed is not the proper type.

Cons:

It does not validate the input as an integer, it only typecasts it to an integer regardless of passed type.
Malformed IDs are “allowed” in that they are converted to 0 (int) value and the script continues as if you requested an object with an ID of 0. (see php docs warnings below).
Only time (int) typecasting throws an exception is when a class is typecast as an (int) and since WordPress does not inherently catch exceptions for validating it is relatively useless.
The program does not know that a malformed ID was passed to the request until after the database query fails or returns an unexpected result. **early detection was cited as a reason for doing this early and throughout the codebase instead of sanitizing just before post injection, however the implications of an unconditional typecast seem to not have been taken into account.
Pros

Sanitizes input to a integer which is the only ‘type’ an ID can take in the database. Anything other than an integer passed to the database as an ID would be unnecessary.
Faster

includes/admin_import.php
kiyote33: May need some help on this as I can't see a work around other than maybe restructuring the entire admin_import.php file. Hesitant to do that! Everything seems to be working well the way it is.

includes/admin_import.php
kiyote33: Dropped the print_r and now simply use tsml_alert($message) for the html table display.

includes/functions.php
kiyote33: Update branding reset to Refresh.

includes/admin_import.php
joshreisner: great! this makes the user feedback consistent with how we report other results like JSON parse errors, etc

includes/functions.php
joshreisner: is the ??? because the instructions aren't clear, or the issue isn't clear? 

the issue is that this is showing up as code you're adding to the branch, but it's already in `main`

i think if you update git it will sync the changes properly

includes/functions.php
joshreisner: ah right, that makes sense. i see we set the last updated timestamp with `current_time('timestamp')` can we use the same method to compare it - that way we are sure we are comparing the same thing

includes/functions.php
joshreisner: as above, `current_time('timestamp')` looks like a better method

includes/admin_import.php
joshreisner: can we use `tsml_alert` here?

includes/functions.php
kiyote33: Sorry, I'm multi-tasking on a zoom with my siblings. I've merged main into my branch and committed some changes into the branch. I'll do another one tomorrow to get the rest of your review items in.

includes/functions.php
kiyote33: Get option has been implemented pretty much anywhere I could find that we are doing datetime formatting.

includes/admin_import.php
kiyote33: Yep. We are using tsml_alert now. I've dropped the print_r.

includes/functions.php
joshreisner: Sure so instead of having both `$db_meetings` and `$all_db_meetings` and the repeat loop, you could do it like this:

```
$db_meetings = array_filter(tsml_get_meetings(), function ($meeting) {
  return in_array($meeting['id'], $ds_ids);
});
```

`array_filter` loops through the array (in this case`tsml_get_meetings`) and uses a callback function to decide whether to filter each item in or out

includes/admin_import.php
kiyote33: Done

includes/admin_import.php
kiyote33: Done

includes/functions.php
kiyote33: Done

includes/functions.php
kiyote33: The array_filter() function filters the values of a single array using the callback function. I'm processing two arrays: get_meetings and ds_ids. So the above code passes the meetings array, but the ds_ids array is undefined.  I did get it to work when I globalized the ds_ids array (which I renamed to $tsml_data_source_ids). The code that worked for me: 	**$db_meetings = array_filter(tsml_get_meetings(), function ($meeting) { global $tsml_data_source_ids;  return in_array($meeting['id'], $tsml_data_source_ids); });**

includes/functions.php
kiyote33: Done

includes/admin_import.php
kiyote33: Done

joshreisner:
this PR is superseded by https://github.com/code4recovery/12-step-meeting-list/pull/1116

PR #1107, opened 2023-03-29
dont try to localize slug

### LINKED ISSUE #1106: meetings page 404 when language is localized
i can recreate the problem locally if i set my site language to russian and save permalinks (with no theme customizations). this is true both in the current version of the site and recent versions. what happens is that the site seems to think that meetings should be at `/собрания` but this address is not found

i can fix it by adding `$tsml_slug = "собрания"` to my theme's functions.php, then the meetings page properly shows up at `/собрания` -- this can be anything, are you doing this too? this would explain why your meetings page is `/schedule` -- this is the exact advice we give people here: https://wordpress.org/plugins/12-step-meeting-list/#can%20i%20change%20the%20url%20of%20the%20meetings%20list%3F

_Originally posted by @joshreisner in https://github.com/code4recovery/12-step-meeting-list/discussions/1095#discussioncomment-5460427_

joshreisner:
fixes #1106 

to recreate this problem locally, set the website language to russian, and save permalinks. you will start seeing 404s on your meetings page. 

then pull this PR down, you will start seeing the page showing up again at `/meetings` (after resaving permalinks)

users can still customize the slug to something else with the normal method of setting `$tsml_slug` in their functions.php - this change will just stop it trying to localize it for them.

### DIFF
--- a/includes/variables.php
+++ b/includes/variables.php
@@ -586,10 +586,10 @@ $tsml_ui_config = [];
 $tsml_days = $tsml_days_order = $tsml_programs = $tsml_types_in_use = $tsml_strings = [];
 
 //string url for the meeting finder, or false for no automatic archive page
-if (!isset($tsml_slug)) $tsml_slug = null;
+if (empty($tsml_slug)) $tsml_slug = 'meetings';
 
 add_action('plugins_loaded', function () {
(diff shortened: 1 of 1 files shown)


kiyote33 (approved):
Looks good to me.

PR #1116, opened 2023-04-15
Wp import data sources redesign

kiyote33:
image

The WordPress maintainable records can now be imported via this new feature: the **Add Meetings** button found on the Direct Entry (local) listing of the "Your Data" card. If you already have records loaded then this new feature turns into either an **Append Meetings** button or a **Replace Meetings** button depending on the **Delete local records** dropdown setting. The new feature's general behavior pretty much mimics the behavior found previously in the CSV Import feature which appended to or replaced the entire non data sourced record set depending on the delete option selected prior to beginning the import. 
Only the external import options appear in the radio button import sub-panel of the "Your Data" card, with their option names being shortened to **CSV Import** and **JSON Import** respectively. 
The **CSV Import** now has a new associated **Upload File** button in the Update column of "Your Data" listing which allows users to simply select the exact same name of their upload file to begin the Update process. 
The **JSON Import** option for updates remains the same with only the title of the button being changed to **Refresh Feed**, better reflecting the difference from the other update operations being listed. 
Both types of external imports do the differential update, where just records in the database which are different from those being imported are actually updated.

This PR has been shelved effective mid-july 2023.


includes/admin_import.php
```diff
@@ -1,15 +1,28 @@
...
 if (!function_exists('tsml_import_page')) {
 
 	function tsml_import_page()
 	{
-		global $tsml_data_sources, $tsml_programs, $tsml_program, $tsml_nonce, $tsml_sharing, $tsml_slug, $tsml_change_detect;
-
+		global $tsml_data_sources, $tsml_programs, $tsml_program, $tsml_nonce, $tsml_sharing, $tsml_slug, $tsml_detection_test_mode;
+		
 		$error = false;
-		$tsml_data_sources = get_option('tsml_data_sources', array());
```
anchovie91471, line 12: SInce $tsml_data_sources is being set here. Shouldn't the `global $tsml_data_sources` be removed?

anchovie91471:
Line 29
`ini_set('auto_detect_line_endings', 1); //to handle mac \r line endings`

This is deprecated and will be removed in PHP 9 so, for forward compatibility, we might consider removing that line. There hasn't been a MacOS that used \r since 2001 so I figure it's safe to remove it.

includes/functions.php
```diff
@@ -1342,7 +1360,8 @@ function tsml_import_buffer_set($meetings, $data_source_url = null, $data_source
 {
 	global $tsml_programs, $tsml_program, $tsml_days, $tsml_meeting_attendance_options, $tsml_data_sources;
 
-	if (strpos($data_source_url, "sheets.googleapis.com") !== false) {
+	//if (strpos($data_source_url, "sheets.googleapis.com") !== false) { refactored to be PHP 8.0 compliant
+	if (str_contains($data_source_url, "sheets.googleapis.com") !== false) {
```
anchovie91471, line 1364: I'm still getting deprecated notices on this line. Changing this line to the following worked to remove the notice:
`if (!empty($data_source_url) && str_contains($data_source_url, "sheets.googleapis.com") !== false) {`

anchovie91471:
When importing a CSV with abbreviated meeting types and spaces between each abbreviation (ex: "C, D, EN, SM"). The first type is the only one that shows up. Types get imported as "C", " D", " EN", " SM", so the ones with spaces don't match the spec.

We should probably change functions.php line 1555 to trim spaces before adding types to the array:
`$meetings[$i]['types'][] = trim($type);`

includes/admin_import.php
```diff
@@ -248,243 +371,280 @@ function tsml_import_page()
...
-									<th class="align-center"><?php _e('Meetings', '12-step-meeting-list') ?></th>
-									<th class="align-right"><?php _e('Last Refresh', '12-step-meeting-list') ?></th>
-									<th class="small"></th>
-								</tr>
-							</thead>
-							<tbody>
-								<?php foreach ($tsml_data_sources as $feed => $properties) { ?>
-									<tr data-source="<?php echo $feed ?>">
-										<td class="small ">
-											<form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
```
anchovie91471, line 307: Mainly cosmetic, but I would wrap the form in a conditional so it doesn't render and can't be un-hidden in dev tools:

`<?php if ($properties['type'] !== 'CSV') { ?>
<form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
												
											<?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
											<input type="hidden" name="tsml_add_data_source" value="<?php echo $feed ?>">
											<input type="hidden" name="tsml_add_data_source_name" value="<?php echo @$properties['name'] ?>">
											<input type="hidden" name="tsml_add_data_source_parent_region_id" value="<?php echo @$properties['parent_region_id'] ?>">
											<?php
												if($properties['type']!=='CSV')	{ ?>
													<input type="submit" value="Refresh" class="button button-small" style="display: block"; > 
											<?php } else { ?>  
													<input type="submit" value="Refresh" class="button button-small" style="display: none"; > 
											<?php } ?>													
										</form>
<?php } ?>`

includes/admin_import.php
```diff
@@ -248,243 +371,280 @@ function tsml_import_page()
...
-									<th class="align-left"><?php _e('Change Detection', '12-step-meeting-list') ?></th>
-									<th class="align-center"><?php _e('Meetings', '12-step-meeting-list') ?></th>
-									<th class="align-right"><?php _e('Last Refresh', '12-step-meeting-list') ?></th>
-									<th class="small"></th>
-								</tr>
-							</thead>
-							<tbody>
-								<?php foreach ($tsml_data_sources as $feed => $properties) { ?>
-									<tr data-source="<?php echo $feed ?>">
-										<td class="small ">
```
anchovie91471, line 306: Mainly cosmetic, but I would wrap the form in a conditional so it doesn't render and can't be un-hidden in dev tools:

```
<?php if ($properties['type'] !== 'CSV') { ?>
	<form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">	
		<?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
		<input type="hidden" name="tsml_add_data_source" value="<?php echo $feed ?>">
		<input type="hidden" name="tsml_add_data_source_name" value="<?php echo @$properties['name'] ?>">
		<input type="hidden" name="tsml_add_data_source_parent_region_id" value="<?php echo @$properties['parent_region_id'] ?>">
		<?php
			if($properties['type']!=='CSV')	{ ?>
				<input type="submit" value="Refresh" class="button button-small" style="display: block"; > 
			<?php } else { ?>  
				<input type="submit" value="Refresh" class="button button-small" style="display: none"; > 
			<?php } ?>													
		</form>
<?php } ?>
```

includes/admin_import.php
```diff
@@ -248,243 +371,280 @@ function tsml_import_page()
...
-												<?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
-												<input type="hidden" name="tsml_add_data_source" value="<?php echo $feed ?>">
-												<input type="hidden" name="tsml_add_data_source_name" value="<?php echo @$properties['name'] ?>">
-												<input type="hidden" name="tsml_add_data_source_parent_region_id" value="<?php echo @$properties['parent_region_id'] ?>">
-												<input type="hidden" name="tsml_add_data_source_change_detect" value="<?php echo @$properties['change_detect'] ?>">
-												<input type="submit" value="Refresh" class="button button-small">
-											</form>
-										</td>
-										<td>
-											<a href="<?php echo $feed ?>" target="_blank">
```
anchovie91471, line 317: Also, just a cosmetic change. The link for CSV doesn't (and presumably shouldn't) open the file.

```
<?php if ($properties['type'] !== 'CSV') { ?>
	<a href="<?php echo $feed ?>" target="_blank">
		<?php echo !empty($properties['name']) ? $properties['name'] : __('Unnamed Feed', '12-step-meeting-list') ?>
	</a>
<?php } else { ?>
	<?php echo ! empty( $properties['name'] ) ? $properties['name'] : __( 'Unnamed Feed', '12-step-meeting-list' ); ?>
<?php } ?>
```

includes/admin_import.php
```diff
@@ -248,243 +371,280 @@ function tsml_import_page()
...
+			tsml_alert(__('Data Sources Debug Mode setting changed.', '12-step-meeting-list'));
+		}
+
+
 		/*debugging
 		delete_option('tsml_data_sources');
 		tsml_delete('everything');
 		tsml_delete_orphans();
 		*/
 ?>
```
anchovie91471, line 395: This is a total hack, but the only thing I've come up with that fixes the issue with the Wordpress admin footer overlapping the page.

```
<style>
	html, body {
		position: relative;
		height: auto;
	}
</style>
```

includes/admin_import.php
```diff
@@ -542,9 +704,37 @@ function tsml_import_page()
...
+									<?php } ?>
+								</select>
+							</form>
+						<?php } ?>
+					</div>
 				</div>
 			</div>
 		</div>
+	</div>
+</div>
```
anchovie91471, line 735: Removing this </div> will fix the issue with the footer overlapping the page.

includes/admin_import.php
```diff
@@ -248,243 +371,280 @@ function tsml_import_page()
...
+								<ul class="types">
+									<?php foreach ($tsml_programs[$tsml_program]['types'] as $value) { ?>
+										<li><?php echo $value ?></li>
 									<?php } ?>
-								</p>
-								<input type="submit" class="button" value="<?php _e('Begin', '12-step-meeting-list') ?>">
-							</form>
-						</div>
+								</ul>
+								</li>
```
anchovie91471, line 621: Unnecessary closing tag. Can be removed.

includes/admin_import.php
kiyote33: Hi @anchovie91471, I come from a MS VB/C# background where we almost never used global variables. Come to think of it, even when I was programming in COBOL on the mainframe we pretty much never used Gobals.  I vaguely remember them from Basic A programs back in the 80's, but that came with lot's of problems, especially when debugging and trying to find where values got changed.  I'm more comfortable keeping variables to a local scope.   I'm not sure what the use of Global variables rules are with PHP, but I assume there is a reason for having them declared that way. Feel free to enlighten me as I'm not real sure what you're getting at.

Thx...

includes/functions.php
kiyote33: This is outside the scope of this PR, but I agree with implementing this fix to get rid of those annoying deprecation notices.

includes/admin_import.php
kiyote33: I'm not completely comfortable with this code. I've never seen it executed, and I haven't done much in the way of trying to load anything but CSV files.

includes/admin_import.php
kiyote33: Thanks. I was thinking we should remove the footer altogether.

includes/admin_import.php
kiyote33: I'm a little confused with which closing tag you are referring to. Which one do you see as unnecessary?

includes/admin_import.php
anchovie91471: I agree completely about the use of globals. It gets especially confusing, at times, to know where a value is coming from when it can change literally anywhere. Hotly contested issue in the PHP community.

As far as the code suggestion, after looking again, I realize you are using it exactly how it's meant. You're declaring it, then setting its value. When I made the suggestion, I was thinking from the standpoint of there being no reason to grab a variable from the global scope and then immediately set it, but obviously that's the whole point so that's my bad.

includes/admin_import.php
anchovie91471: Line 621 </li>.

kiyote33: Hilarious & Obvious. Why would a closing line item be outside the closing unordered list tag? It's gone now!
'

includes/admin_import.php
anchovie91471: When I add a CSV and it shows up in the feed list, currently the feed name is a link. When clicked, it opens a new browser window with just "meeetings.csv" in the address bar and of course that throws a 404. I looked to see if you were saving the uploaded csv on the server somewhere (like in wp-content/uploads) but couldn't find the file so I assume you're removing it after processing? I just figured removing the link for csv uploads would be the way to go, but it you want people to be able to open the csv and review it, then this isn't the way to go.

Underearners-Anonymous:
Please add Underearners Anonymous to the 12-step program options on the 12-step meeting list app. Thanks

joshreisner:
closing this previous legacy PR to avoid confusion - it can be reopened at any time if i'm in error

PR #1124, opened 2023-05-02
Add SPAD format to NA program

pjaudiomv:
With the release of the A Spiritual Principle a Day book, NAWS has created a new SPAD format and some of our service bodies who have groups use this format. This will allow us to us it in TSML as well.


joshreisner (approved):
nice

Issue #1049, opened 2023-01-13
Problem with Bulk Action remove meetings


joshreisner:
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1048

<div type='discussions-op-text'>

<sup>Originally posted by **brittbuhl** January 13, 2023</sup>
Hello, my name is Britt and I am new to using TSML and have noticed an issue that I wanted to see if anyone else had noticed.  

I had entered our entire District meeting list and noticed some duplicates so I want to the meetings list and filtered the meetings by region to find the duplicate meetings and remove them.  I found the meetings I wanted to remove and checked them.  

Then I selected the bulk action "Move to Trash" option and clicked apply.  Once I did this, my entire meeting list disappeared from the website.

I then went to the trash area and found the meetings there, selected them all and clicked "restore".  The meetings then moved to a "drafts" category (not published) and, lo and behold, my entire meeting list re-appeared on the website.  I left the duplicate meetings in the "drafts" area and they don't appear on the website so - mission accomplished.  But, I noticed that error with the bulk action button and wanted to see if anyone else had noticed this?

Note:  when I remove a meeting individually (send it to the trash folder) - this doesn't cause my entire meeting list to disappear.  It seems that only using the bulk action button does that.

Anyone else notice this issue?  Any suggestions?

Thanks a lot!

</div>

joshreisner:
this was fixed by @anchovie91471 last month in https://github.com/code4recovery/12-step-meeting-list/pull/1129

PR #1129, opened 2023-05-15
Fix for #1049 - Problem with Bulk Action remove meetings

anchovie91471:
Closing #1049 

The important parts of this fix are:
    1. The addition of the "transition_post_status" action hook, that refrehes the tsml_cache when post statuses are changed (published -> trash, draft -> published, etc.)
    2. The $wpdb->is_main_query() check in the "pre_get_posts" action that prevents the filters form being applied to subsequent queries (this was being applied to the query in the get_meetings() function, causing the tsml_cahce json file to be empty.

Details:
- Added Wordpress "transition_post_status" action that runs tsml_cache_rebuild() anytime meeting post statuses are changed
- "restrict_manage_posts" action - Changed to use built-in $post_type argument - Removed values in wp_dropdown_categories() that were identical to defaults - Sanitize incoming "region" $_GET variable before setting value in dropdown
- "parse_query" action - Changed to "pre_get_posts" (newer and better documented action) - Changed to use global $post_type argument instead of $_GET - Added $query->is_main_query() check before setting filters. - Refectored $parent_ids query to use $wpdb->prepare() to prevent SQL injection - Sanitize incoming "region" $_GET variable before using in $wpdb->get_col()

### DIFF
--- a/includes/admin_lists.php
+++ b/includes/admin_lists.php
@@ -1,41 +1,46 @@
 <?php
 
-# Adding region dropdown to filter
-add_action('restrict_manage_posts', function () {
+// Rebuild tsml cache when using bulk actions (edit status or trash)
+add_action('transition_post_status', function ($new_status, $old_status, $post) {
+    if ($post->post_type === 'tsml_meeting') {
+        tsml_cache_rebuild();
+    }
+}, 10, 3);
 
-    if (@$_GET['post_type'] != 'tsml_meeting') {
+# Adding region dropdown to filter
+add_action('restrict_manage_posts', function ($post_type) {
+
+    if( $post_type !== 'tsml_meeting' ) {
         return;
     }
 
     wp_dropdown_categories([
         'show_option_all' => 'Region',
         'orderby' => 'tax_name',
-        'hide_empty' => true,
-        'selected' => !empty($_GET['region']) ? $_GET['region'] : null,
+        'selected' => !empty($_GET['region']) ? sanitize_text_field($_GET['region']) : '',
         'hierarchical' => true,
         'name' => 'region',
         'taxonomy' => 'tsml_region',
         'hide_if_empty' => true,
-        'value_field' => 'term_id',
     ]);
-});
+}, 10, 1);
 
 # If filter is set, restrict results
 add_filter(
-    'parse_query',
+    'pre_get_posts',
     function ($query) {
-        global $pagenow, $wpdb;
+        global $post_type, $pagenow, $wpdb;
 
-        if (($pagenow != 'edit.php') || (@$_GET['post_type'] != 'tsml_meeting') || empty($_GET['region'])) {
-            return;
+        if(!empty($_GET['region']) && $pagenow === 'edit.php' && $post_type === 'tsml_meeting' && $query->is_main_query()) {
+            $parent_ids = $wpdb->get_col($wpdb->prepare(
+                "SELECT p.ID FROM $wpdb->posts p 
+	                    JOIN $wpdb->term_relationships r ON r.object_id = p.ID 
+	                    JOIN $wpdb->term_taxonomy x ON x.term_taxonomy_id = r.term_taxonomy_id 
+	                    WHERE x.term_id = %d",
+                    intval(sanitize_text_field($_GET['region']))
+            ));
+            $query->query_vars['post_parent__in'] = empty($parent_ids) ? [0] : $parent_ids;
         }
-
-        $parent_ids = $wpdb->get_col('SELECT p.ID FROM ' . $wpdb->posts . ' p
-            JOIN ' . $wpdb->term_relationships . ' r ON r.object_id = p.ID
-            JOIN ' . $wpdb->term_taxonomy . ' x ON x.term_taxonomy_id = r.term_taxonomy_id
-            WHERE x.term_id = ' . intval($_GET['region']));
-
-        $query->query_vars['post_parent__in'] = empty($parent_ids) ? [0] : $parent_ids;
     }
 );
 
--- a/includes/admin_lists.php
+++ b/includes/admin_lists.php
@@ -1,16 +1,9 @@
 <?php
 
-// Rebuild tsml cache when using bulk actions (edit status or trash)
-add_action('transition_post_status', function ($new_status, $old_status, $post) {
-    if ($post->post_type === 'tsml_meeting') {
-        tsml_cache_rebuild();
-    }
-}, 10, 3);
-
 # Adding region dropdown to filter
 add_action('restrict_manage_posts', function ($post_type) {
 
-    if( $post_type !== 'tsml_meeting' ) {
+    if($post_type !== 'tsml_meeting') {
         return;
     }
 
--- a/includes/init.php
+++ b/includes/init.php
@@ -82,11 +82,12 @@ add_action('init', function () {
 });
 
 if (is_admin()) {
-    // Rebuild tsml cache when using bulk actions (edit status or trash)
+    // Rebuild tsml cache when using bulk actions to edit status.
+    // Uses a transient to keep track the array of post ids that are being processed
     add_action('transition_post_status', function ($new_status, $old_status, $post) {
         global $pagenow;
         if ($post->post_type === 'tsml_meeting' && $new_status === 'publish' && $pagenow !== 'post.php' && $pagenow !== 'admin-ajax.php') {
-            // Look for transient first. Get it from $_GET if it doesn't exist
+            // Try to load transient first or $_GET['post'] if it doesn't exist
             if ($bulk_ids = get_transient('tsml_bulk_process')) {
                 // Remove current post ID from array
                 if (($key = array_search($post->ID, $bulk_ids)) !== false) {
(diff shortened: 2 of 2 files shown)


includes/admin_lists.php
```diff
@@ -1,41 +1,46 @@
 <?php
 
+// Rebuild tsml cache when using bulk actions (edit status or trash)
+add_action('transition_post_status', function ($new_status, $old_status, $post) {
```
joshreisner, line 4: how many times does this fire when performing bulk actions? would it fire once or once for every post being changed?

are there places (eg in init.php) that we should be taking tsml_cache_rebuild out since it will be caught here?

seems like this might fire whenever there's a post update not just a status update (which might not be bad?) do we want to restrict it with an if statement or no? https://developer.wordpress.org/reference/hooks/transition_post_status/#comment-2977

includes/admin_lists.php
```diff
@@ -1,41 +1,46 @@
...
+        tsml_cache_rebuild();
+    }
+}, 10, 3);
+
 # Adding region dropdown to filter
-add_action('restrict_manage_posts', function () {
+add_action('restrict_manage_posts', function ($post_type) {
 
-    if (@$_GET['post_type'] != 'tsml_meeting') {
+    if( $post_type !== 'tsml_meeting' ) {
```
joshreisner, line 13: would be great to get this PSR12 formatted with https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client - otherwise there will be unrelated whitespace changes on the next PR

includes/admin_lists.php
```diff
@@ -1,41 +1,46 @@
...
         return;
     }
 
     wp_dropdown_categories([
         'show_option_all' => 'Region',
         'orderby' => 'tax_name',
-        'hide_empty' => true,
-        'selected' => !empty($_GET['region']) ? $_GET['region'] : null,
+        'selected' => !empty($_GET['region']) ? sanitize_text_field($_GET['region']) : '',
```
joshreisner, line 20: good catch

joshreisner:
ohhhhh! so the cache was empty when deleting a region because the region was still set (to a newly deleted region) when rebuilding the cache? wow cool, great work

includes/admin_lists.php
anchovie91471: This is a very good point that I didn't think of. I believe it will run once for each meeting, anytime the status changes (regardless of if it changes on the edit meeting page or in bulk actions or anywhere else). I also think it runs when a meeting is saved, whether the status ACTUALLY changes or not, so I need to at least change line 5 to: 

`if ($post->post_type === 'tsml_meeting' && $old_status !== $new_status) {`

But I don't think that's really enough. I'll see where else we're rebuilding and report back. We may be able to remove some rebuilds elsewhere of just put the rebuild in an existing hook.

anchovie91471: I've pulled together some info so we can discuss where to call tsml_cache_rebuild() to ensure it's called only when needed.

Here is a google sheet listing everywhere it's called, including the file/purpose: https://docs.google.com/spreadsheets/d/1Py9oQ2NsGvGLR-FrqB2NZr6cpNamQW3lQXSP1a41_8w/edit#gid=0

I also ran several operations on my dev site and logged where and how many times cache rebuild was called. Below is the log entries for that. The bold all caps text below each group of log entries explains what I did in Wordpress to trigger the entries. Each log entry is paired. 1 entry that indicates which hook the rebuild was run under and the 2nd is just "tsml cache rebuilt" to indicate it was successful.

[15-May-2023 18:09:16 UTC] transition_post_status
[15-May-2023 18:09:16 UTC] tsml cache rebuilt
[15-May-2023 18:09:16 UTC] trashed_post/untrashed_post
[15-May-2023 18:09:16 UTC] tsml cache rebuilt
[15-May-2023 18:09:16 UTC] transition_post_status
[15-May-2023 18:09:16 UTC] tsml cache rebuilt
[15-May-2023 18:09:16 UTC] trashed_post/untrashed_post
[15-May-2023 18:09:16 UTC] tsml cache rebuilt
**BULK ACTION MOVE 2 MEETINGS TO TRASH**

[15-May-2023 18:09:52 UTC] transition_post_status
[15-May-2023 18:09:52 UTC] tsml cache rebuilt
[15-May-2023 18:09:52 UTC] trashed_post/untrashed_post
[15-May-2023 18:09:52 UTC] tsml cache rebuilt
[15-May-2023 18:09:52 UTC] transition_post_status
[15-May-2023 18:09:52 UTC] tsml cache rebuilt
[15-May-2023 18:09:52 UTC] trashed_post/untrashed_post
[15-May-2023 18:09:52 UTC] tsml cache rebuilt
**BULK ACTION RESTORE 2 MEETINGS**

[15-May-2023 18:10:28 UTC] transition_post_status
[15-May-2023 18:10:28 UTC] tsml cache rebuilt
[15-May-2023 18:10:28 UTC] transition_post_status
[15-May-2023 18:10:28 UTC] tsml cache rebuilt
**BULK ACTION EDIT 2 MEETINGS CHANGE TO PUBLISHED**

[15-May-2023 18:11:23 UTC] transition_post_status
[15-May-2023 18:11:23 UTC] tsml cache rebuilt
[15-May-2023 18:11:23 UTC] trashed_post/untrashed_post
[15-May-2023 18:11:23 UTC] tsml cache rebuilt
**TRASH 1 MEETING FROM MEETING LISTING EDIT PAGE**

[15-May-2023 18:12:18 UTC] transition_post_status
[15-May-2023 18:12:18 UTC] tsml cache rebuilt
[15-May-2023 18:12:18 UTC] trashed_post/untrashed_post
[15-May-2023 18:12:18 UTC] tsml cache rebuilt
**RESTORE 1 MEETING FROM MEETING LISTING TRASH PAGE**

[15-May-2023 18:13:02 UTC] transition_post_status
[15-May-2023 18:13:02 UTC] tsml cache rebuilt
[15-May-2023 18:13:02 UTC] save_post
[15-May-2023 18:13:02 UTC] tsml cache rebuilt
**PUBLISH 1 MEETING FROM THE MEETING'S EDIT PAGE (NO OTHER MEETING DATA CHANGED)**

[15-May-2023 18:14:04 UTC] save_post
[15-May-2023 18:14:04 UTC] tsml cache rebuilt
**UPDATE 1 MEETING FROM THE MEETING'S EDIT PAGE**

[15-May-2023 18:19:17 UTC] transition_post_status
[15-May-2023 18:19:17 UTC] tsml cache rebuilt
[15-May-2023 18:19:17 UTC] wp_ajax_tsml_import
[15-May-2023 18:19:17 UTC] tsml cache rebuilt
**IMPORT 1 MEETING (CSV)**

[15-May-2023 18:32:59 UTC] transition_post_status
[15-May-2023 18:32:59 UTC] tsml cache rebuilt
[15-May-2023 18:32:59 UTC] wp_ajax_tsml_import
[15-May-2023 18:32:59 UTC] tsml cache rebuilt
**IMPORT 1 MEETING (JSON)**

[15-May-2023 19:13:14 UTC] tsml_delete
[15-May-2023 19:13:14 UTC] tsml cache rebuilt
**PERMANENTLY DELETE 1 MEETING**

I realize we probably don't want to rebuild multiple times if we're doing a single operation (i.e. Moving 2 meetings to trash). We can discuss options to reduce this. For now, I think we should decide where to keep the rebuilds and where to remove them.

My opinion:

1. Since save_post is working fine and multiple rebuilds are not triggered, I say we leave that one as-is
2. Since tsml_delete is working fine and multiple rebuilds are not triggered, I say we leave that one as-is also
3. transition_post_status (added in this PR) is running under every other condition. I say remove the call to tsml_cache_rebuild() from: "trashed_post/untrashed_post" & "wp_ajax_tsml_import"

This will leave us with 1 scenario where the rebuild will run twice: **PUBLISH 1 MEETING FROM THE MEETING'S EDIT PAGE (NO OTHER MEETING DATA CHANGED)**. I can explore the possibility of a conditional that will limit the rebuild under this scenario.

Let me know what you think!

anchovie91471:
> 

So what was actually happening is the code that added the query filter to reduce the list of meetings to just the selected region, was carrying over to the tsml_get_meeting() function (which is where the cache file is created). I didn't dive super-deep into that function but the face that it was a reduced number of meetings (only ones from whatever region ID was selected), along with the other conditionals in that function were effectively resulting in 0 meetings being cached. The file was being rebuild but it just contained and empty json array `[]`

joshreisner:
yes, makes perfect sense now thanks

includes/admin_lists.php
joshreisner: ok thanks for the audit! couple ideas:

- we might do well to move this to init.php and essentially replace the `tsml_trash_change` stuff with this, since it's global in effect and does the same thing but in more cases
- we should call `tsml_update_types_in_use()` to the hook since it might need to be updated after removing some posts
- we could add a global variable called `$tsml_bulk_change_in_progress` and use it to prevent excessive re-firing of this hook during bulk actions, then we could keep it in the import (and just set it to `true` at the beginning and `false` at the end of the import

anchovie91471: Latest push contains all 3. Use WP transients instead of the global, but accomplishes the same thing.

joshreisner (approved):
🥇

PR #1131, opened 2023-05-16
partly revert slug change

joshreisner:
this change should still prevent the slug localization problems while going back to the 'old' way of defaulting to 'meetings' when `$tsml_slug` is not specified locally

### DIFF
--- a/includes/variables.php
+++ b/includes/variables.php
@@ -586,10 +586,10 @@ $tsml_ui_config = [];
 $tsml_days = $tsml_days_order = $tsml_programs = $tsml_types_in_use = $tsml_strings = [];
 
 //string url for the meeting finder, or false for no automatic archive page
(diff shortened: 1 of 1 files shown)


joshreisner:
got a [report of an error](https://github.com/code4recovery/12-step-meeting-list/discussions/1130#discussioncomment-5916579), not able to reproduce

kiyote33 (approved):
Interesting got this error when attemption to load from a local dev json file:             [http_request_failed] => Array
                (
                    [0] => cURL error 60: SSL certificate problem: self signed certificate
                )
but it loaded fine from my public dev site at eco.parkercanada.info. Yesterday I had no problem loading from the same local json file. Must be something different in the way I configured the pristine test environments.

joshreisner:
thanks - that looks like an unrelated issue with the SSL certfiicate

PR #1201, opened 2023-08-25
add survivors of incest anonymous

joshreisner:
for https://github.com/code4recovery/12-step-meeting-list/discussions/1196


kiyote33 (commented):
We have major changes scheduled for the 3.15 milestone. This seems like a rather trivial update that should have gone under 3.14.20. Is it intended to drop the 3.15 milestones or reset them to 3.16? What's the plan?

joshreisner:
@kiyote33 ok i will update it to be 3.14.20

eventually i think we should start using more `minor` versions and reserve `patch` versions just for bug fixes 

from https://semver.org

> Given a version number MAJOR.MINOR.PATCH, increment the:
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes

kiyote33 (approved):
Thanks. We're good to go.

PR #1203, opened 2023-08-28
fix maps javascript errors

joshreisner:
fixes two maps javascript errors that can occur if values are undefined, reported in https://github.com/code4recovery/12-step-meeting-list/discussions/1202


kiyote33 (approved):
Works fine with/without location and region. Ready to go...

PR #1211, opened 2023-09-02
fix: omit regions array from JSON feed when empty

### LINKED ISSUE #1210: when region is not set, regions should not be [ null ]
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1209

when a region is not set, the `regions` key is `[null]` in the json as pictured below. instead the key should simply not exist, which is what the `region` key currently does

Screenshot 2023-09-01 at 1 22 47 PM

fwiw i'm 99% sure i caused this originally

joshreisner:
fixes #1210 

when there are no regions, don't add the key `"regions": [ null ],` to the JSON feed


kiyote33 (approved):
Successful test: Regions key is now removed from JSON when Region removed from a meeting. Good to go...

PR #1240, opened 2023-10-22
Fix ACF compatibility

joshreisner:
check out my comment here: https://github.com/code4recovery/12-step-meeting-list/discussions/1239?converting=1#discussioncomment-7351769

to reproduce the issue:

* install Advanced Custom Fields on a site that has TSML installed.
* now add create a field group
* add a time field to the field group
* say that the field group should only display on posts
* create a new post and click on the the field group, it will look like this if this branch is active
Screenshot 2023-10-22 at 8 56 30 AM

however `main` looks like this
Screenshot 2023-10-22 at 8 57 04 AM

hopefully no other issues are observed

### DIFF
--- a/12-step-meeting-list.php
+++ b/12-step-meeting-list.php
@@ -4,7 +4,7 @@
  * Plugin Name: 12 Step Meeting List
  * Plugin URI: https://wordpress.org/plugins/12-step-meeting-list/
  * Description: Manage a list of recovery meetings
- * Version: 3.14.22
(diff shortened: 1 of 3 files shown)


kiyote33 (approved):
Installed fix in test environment with the ACF plugin active. Test page included custom ACF field group with a Date/Time picker. Presentation of the field match what I expected: October 27, 2023 7:00 am. Version updates are good so I think we are good to go.

PR #1249, opened 2023-10-30
Security update

joshreisner:
Based on reported issue - need to confirm


kiyote33 (approved):
Research indicates safe version of wp_remote_get () was created for the purpose of preventing Server Side Request Forgery (SSRF) attacks. So this PR addresses that vulnerability. It tests out OK, invalidating internal host requests like http://127.0.0.1/...! 

Thanks Josh

PR #1277, opened 2023-12-04
fixing issues with wordpress.com

joshreisner:
wordpress.com users are experiencing issues saving meetings

https://github.com/code4recovery/12-step-meeting-list/discussions/1257
https://github.com/code4recovery/12-step-meeting-list/discussions/1273
https://github.com/code4recovery/12-step-meeting-list/discussions/1276

this appears to be because they are using a persistent object cache which we bypass with `$wpdb` functions in certain places (like when setting the parent on a saved meeting). this PR fixes that by clearing the cache after making those types of changes

additionally this PR:

* retires the `tsml_ensure_location_approximate_set` function
* fixes padding on the "Enable Maps on Your Site" notice
* fixes the "Add New Post" label in the sidebar
* adds a function `tsml_dd` for debugging


joshreisner:
updated code per conversation with dave a to use wp_cache_clear() instead of bypassing wp get_post / get_posts functions

Issue #1200, opened 2023-08-24
Meeting Guide Page empty after deactivating, reactivating plugin


joshreisner:
should see if there's something we can do about this

### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1199

<div type='discussions-op-text'>

<sup>Originally posted by **Area18Webservant** August 23, 2023</sup>
Hi all, 

Was having trouble with our new gravity forms plugin, and as one of the initial checks. I tried deactivating plugins one at a time to find any potential contamination. When I have done this before, I usually skip over the 12 step meeting guide plugin, but I accidentally disabled it this time. I enabled it again right away, but now the Meeting guide page is empty. How do I restore my meeting database on that page?</div>

joshreisner:
looks like this is fixed by https://github.com/code4recovery/12-step-meeting-list/pull/1291 - thanks @gkovats !!

PR #1291, opened 2023-12-25
[Issue-1200] Fix rewrite rules after plugin activation / decativation

gkovats:
## Fix for issue #1200 

Used a `rewrite-rules-inspector` to look at the rewrite rules before and after plugin activation, and noticed

1. After plugin activation as second time (not first time), rules look invalid. 

I don't know exactly why they're bad, we register the post types and taxonomies and flush rewrite rules. I tried flushing rewrite rules later, at shutdown, but no luck. Since it seems fine if you flush rewrite rules on a later load, figure it's something with timing. Activation hook happens maybe too later for rewrite / post_type and taxonomy registration to matter. 

To fix, on activation, I set a `tsml_plugin_new_activation` to `true` (`1`), then init checks if truthy, flushes rewrite, sets false, so basically flush rewrite rules on first full load. Works on my end, but I haven't tested on multi site. _Should_ be good.

2. After plugin deactivation, rewrite rules for our plugin still exist.

Not a huge issue, but "leave no trace", figured better to unregister post types and taxonomies before final flush rewrite. Fixes the remnant rules.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -156,15 +156,15 @@ function tsml_plugin_activation()
 {
 	tsml_custom_post_types();
 	flush_rewrite_rules();
-    add_option( 'tsml_plugin_new_activation', true );
-    update_option( 'tsml_plugin_new_activation', true );
+	add_option( 'tsml_plugin_new_activation', true );
+	update_option( 'tsml_plugin_new_activation', true );
 }
 
 //called by register_deactivation_hook in 12-step-meeting-list.php
 //hands off to tsml_custom_post_types
 function tsml_plugin_deactivation()
 {
-    tsml_unregister_custom_post_types();
+	tsml_unregister_custom_post_types();
 	flush_rewrite_rules();
 }
 
@@ -348,9 +348,9 @@ function tsml_custom_post_types()
 //used: 	plugin deactivation, before final rewrite flush
 function tsml_unregister_custom_post_types()
 {
-    unregister_taxonomy('tsml_region');
-    unregister_taxonomy('tsml_location');
-    unregister_taxonomy('tsml_district');
+	unregister_taxonomy('tsml_region');
+	unregister_taxonomy('tsml_location');
+	unregister_taxonomy('tsml_district');
 	unregister_post_type('tsml_meeting');
 	unregister_post_type('tsml_location');
 	unregister_post_type('tsml_group');
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -161,10 +161,27 @@ function tsml_plugin_activation()
 }
 
 //called by register_deactivation_hook in 12-step-meeting-list.php
-//hands off to tsml_custom_post_types
+//clean up custom taxonomies / post types and flush rewrite rules
 function tsml_plugin_deactivation()
 {
-	tsml_unregister_custom_post_types();
+	if ( taxonomy_exists('tsml_region') ) {
+        unregister_taxonomy('tsml_region');
+    }
+	if ( taxonomy_exists('tsml_location') ) {
+	    unregister_taxonomy('tsml_location');
+    }
+	if ( taxonomy_exists('tsml_district') ) {
+	    unregister_taxonomy('tsml_district');
+    }
+    if ( post_type_exists('tsml_meeting') ) {
+	    unregister_post_type('tsml_meeting');
+    }
+    if ( post_type_exists('tsml_location') ) {
+	    unregister_post_type('tsml_location');
+    }
+    if ( post_type_exists('tsml_group') ) {
+	    unregister_post_type('tsml_group');
+    }
 	flush_rewrite_rules();
 }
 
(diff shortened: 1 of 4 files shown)


includes/functions.php
```diff
@@ -152,10 +152,20 @@ function tsml_calculate_attendance_option($types, $approximate)
 
 //called by register_activation_hook in 12-step-meeting-list.php
 //hands off to tsml_custom_post_types
-function tsml_change_activation_state()
+function tsml_plugin_activation()
 {
 	tsml_custom_post_types();
 	flush_rewrite_rules();
+    add_option( 'tsml_plugin_new_activation', true );
+    update_option( 'tsml_plugin_new_activation', true );
```
gkovats, line 160: Maybe I'm doing this wrong, but if an option doesn't exist, you need to add it, and if it does you have to update it. Easier to do both then check and do one or the other.

includes/functions.php
```diff
@@ -2119,7 +2141,7 @@ function tsml_import_changes($feed_meetings, $data_source_url, $data_source_last
 		__('Saturday', '12-step-meeting-list'),
 	];
 
-	// get local meetings 
+	// get local meetings
```
gkovats, line 2144: sorry, trimming trail whitespace on save.

includes/functions.php
```diff
@@ -152,10 +152,20 @@ function tsml_calculate_attendance_option($types, $approximate)
 
 //called by register_activation_hook in 12-step-meeting-list.php
 //hands off to tsml_custom_post_types
-function tsml_change_activation_state()
+function tsml_plugin_activation()
 {
 	tsml_custom_post_types();
 	flush_rewrite_rules();
+	add_option( 'tsml_plugin_new_activation', true );
+	update_option( 'tsml_plugin_new_activation', true );
```
joshreisner, line 160: couple ideas here, but the main one is: if the point of setting the option is to wait to flush the rewrite rules in the init hook, how about instead running `tsml_custom_post_types` above the flush_rewrite_rules here? then we don't need the option or the change to init…

includes/functions.php
```diff
@@ -334,6 +344,18 @@ function tsml_custom_post_types()
...
+//used: 	plugin deactivation, before final rewrite flush
+function tsml_unregister_custom_post_types()
+{
+	unregister_taxonomy('tsml_region');
+	unregister_taxonomy('tsml_location');
+	unregister_taxonomy('tsml_district');
+	unregister_post_type('tsml_meeting');
+	unregister_post_type('tsml_location');
+	unregister_post_type('tsml_group');
+}
```
joshreisner, line 357: feel like it would be cleaner to merge this with `tsml_plugin_deactivation` instead of creating a new function, since it's only ever called from there

joshreisner:
thanks! makes a ton of sense to me now that the problem was how we flush_rewrite_rules() on plugin activation / deactivation, but it might not work if the custom post types were not set then

by the same token though, i think maybe we should check to make sure they exist before unregistering them in the deactivation hook to prevent errors.

includes/functions.php
joshreisner: i think we should also check `post_type_exists` / `taxonomy_exists` though to prevent errors

gkovats: Sounds good, will move.

includes/functions.php
gkovats: Gotcha was the `tsml_custom_post_types()` was already running before the `flush_rewrite_rules()`, so changing order here doesn't seem to have an impact. There's still something up with the timing, when the `activate_${plugin}` plugin action happens that the rewrite rules get picked up, but they appear broken or invalid. I started with this approach, "well, just flush rewrite on the next full page load", but yeah, it's not ideal having an option for this one step.

Will keep digging if there's a way to have it on activation all rewrite rules are good by the end of that plugin activation page request.

gkovats:
@joshreisner Problem was the valuables normally loaded from `variables.php` from `plugins_loaded` action _aren't_ in an activation, since that's happening well after the action. So just triggering them by name now from activation. You'll see I made the anonymous function in variables now `tsml_load_config()`. Holler if there's a better name, but it just needed the `$tsml_slug` value set there. Best all values load as they normally do for standard plugin loaded, in case we expect others later in custom_post_types().

includes/variables.php
```diff
@@ -588,7 +588,8 @@
 //string url for the meeting finder, or false for no automatic archive page
 if (!isset($tsml_slug)) $tsml_slug = null;
```
gkovats, line 590: Here I'm just naming this anonymous function so it can be called below. The core issue was $tsml_slug wasn't set during first call to `tsml_custom_post_types()`

includes/variables.php
```diff
@@ -1418,4 +1419,10 @@
 
 	$tsml_types_in_use = get_option('tsml_types_in_use', []);
 	if (!is_array($tsml_types_in_use)) $tsml_types_in_use = [];
-});
+};
+
+add_action('plugins_loaded', 'tsml_load_config');
```
gkovats, line 1424: below, just seeing if ^ this hook will never trigger, than call it now.
- don't love the function name, glad to rename
- can move ^ into an else clause if you prefer

joshreisner (approved):
ok thanks!

PR #1309, opened 2024-01-15
Whitespace changes from php formatter

joshreisner:
went through our PHP files and formatted them with https://www.devsense.com/en DevSense PHP Tools. hopefully this will help us get the whitespace changes in #1282 out of the diff

no functionality changes as part of this change.

mainly this just standardizes us on spaces instead of tabs, four spaces per level of indentation (per PSR-12)


kiyote33 (approved):
Tested OK.

Issue #1321, opened 2024-01-27
Security Audit


joshreisner:
We have received an unconfirmed report of a possible security issue in TSML. While we wait for details on what that is, there are a number of improvement actions that we can take.

1. we can be sure to verify nonces on every POST request, currently I see dozens of places where we verify nonces, but at least one where we do not: on the import & settings page when removing a data source
2. we can explicitly check for required permissions on any non-public ajax method (i see four)
3. tsml_import_page and tsml_settings page are currently functions which are currently only called from `add_submenu_page` - but since they are functions they technically could be reused. seems like a stretch but we could make them closures or explicitly re-check permissions

kiyote33:
This issue closed with the release of #3.14.27 and #3.14.28

PR #1322, opened 2024-01-27
Add notice, fix some potential security edge-cases

joshreisner:
for #1321

* define constants (so they can't be altered) for permission levels
* create shared functions to require correct permissions or die
* require permission whenever some write action could be taken
* use closures in a few cases where feasible because they are not available to other plugins / themes
* remove database upgrade functions since they are no longer needed and could appear to a security researcher like a vulnerability


anchovie91471 (approved):
Simple enough

kiyote33 (approved):
Interesting that security testing with this fix produced exactly the same result as before: "Sorry, you are not allowed to access this page." for all our menu items, except Regions and Districts which had the "You need a higher level of permission.  Sorry, you are not allowed to manage terms in this taxonomy." I commented out the  "tsml_require_meetings_permission();" code on the import  file with no effect. Both the subscriber and admin roles continued to work as expected. It's looking to me like this code is not having an overriding effect. We have the same state on all my tests as before.  Hopefully it will suffice to have the critical security issue withdrawn by the PatchStack folks. I think we could drop the TODO comment in admin_import and consider this done. Thanks for your work on this Josh...

PR #1335, opened 2024-02-02
enable exporting custom fields

joshreisner:
for https://github.com/code4recovery/12-step-meeting-list/discussions/1334

add the field to ACF
Screenshot 2024-02-02 at 8 17 45 AM

add a value to a meeting
Screenshot 2024-02-02 at 8 17 27 AM

add an array to your theme
```
$tsml_custom_meeting_fields = [
	'test' => 'Test',
];
```

click 'export csv'
you should see your custom fields appended as columns at the end of the CSV ✨ 

(i also noticed a few warnings when saving a meeting, so i added some conditionals to save.php to suppress them)

additionally i've added a more stringent check on CSV permissions

### DIFF
--- a/12-step-meeting-list.php
+++ b/12-step-meeting-list.php
@@ -4,7 +4,7 @@
  * Plugin Name: 12 Step Meeting List
  * Plugin URI: https://wordpress.org/plugins/12-step-meeting-list/
  * Description: Manage a list of recovery meetings
- * Version: 3.14.28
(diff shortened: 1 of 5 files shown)


kiyote33 (approved):
Tested out ok. Might want to update the readme "Tested up to:  6.4.3 " .  Thanks  for the nice fix for discussion #1334.

PR #1336, opened 2024-02-04
clean up security checks and sync fields

joshreisner:
now that CSV exports are restricted to editors and above, we can remove the checks we used to have to prevent exposing contact info

this means we can clean up our code by using the existing `$tsml_contact_fields` global variable. confirmed in testing that:
* contact info is visible in CSV
* contact info only in JSON and visible on meetings page if contacts are public

additionally, this PR cleans up AJAX requests on the meeting list page for legacy UI. previously, when a user landed on the meeting directory, three JSON requests would be fired off: 
1. admin-ajax.php?action=tsml_regions
2. admin-ajax.php?action=tsml_locations
3. admin-ajax.php?action=tsml_groups
the latter two of these were shared with the meeting edit page, creating an unclear security situation.

now those two AJAX endpoints are locked down, and the third `tsml_regions` has been changed to `tsml_typehead` and handles the typeahead data for all three. this makes the frontend more efficient with a single request that only fetches the (public) info it needs

cleans up a few php warnings i encountered when testing with no regions / one region

makes it possible to unset the meeting's region

fixes spacing around bolded text in legacy typeahead


kiyote33 (changes requested):
Works as described. I believe the Contact Visibility on the Settings page defaults to 'Public'. Could we change the default to 'Private'? That would mean an explicit action would need to be taken before someone accidentally breaks the anonymity tradition.  Might it also be wise to  add an anonymity reminder when choosing to make Visibility = Public?

joshreisner:
thanks @kiyote33 it defaults to private: https://github.com/code4recovery/12-step-meeting-list/blob/main/includes/variables.php#L57 (always has)

it's not against traditions to make them public though. don't think a warning is needed

kiyote33 (approved):

PR #1338, opened 2024-02-05
Fix headers and footers in block themes

### LINKED ISSUE #1260: Fix headers and footers in WP block themes
review plugin support for [wordpress block themes](https://developer.wordpress.org/themes/block-themes/)

we may be lacking some support in this area, see:

1. https://discourse.tiaa-forum.org/t/meeting-guide-in-wp-2023-theme/8917
2. https://github.com/code4recovery/12-step-meeting-list/discussions/1258#discussioncomment-7549326

actions:
* review the plugin display in legacy and TSML UI mode 
* determine what should be added to our [template](https://github.com/code4recovery/12-step-meeting-list/blob/main/templates/archive-meetings.php#L312-L319) to make block themes such as Twenty Twenty Three and Furnishop render properly

joshreisner:
fix #1260 

before:
before

after:
after

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2388,3 +2388,22 @@ function tsml_date_localised($format, $timestamp = null)
     return $datetime->format($format);
 }
 /* ******************** end of data_source_change_detection ******************** */
+
+function tsml_header()
+{
+    if (function_exists('wp_is_block_theme') && wp_is_block_theme()) {
+        include(TSML_PATH . '/templates/header.php');
+    } else {
+        get_header();
+    }
+}
+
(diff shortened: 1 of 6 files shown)


kiyote33 (approved):
Struggled to right align the menu. Would be nice if that was the default. Other than that it's Great!!! 

FYI: Tested with Twenty Two, Three, and Four. Looks good.

joshreisner:
yeah i tried to figure that out too! looks like it's not outputting all the needed CSS - not sure what, if anything, is missing that would output that. might be a bug in twentytwentyfour?

Issue #1342, opened 2024-02-08
Error deleting plugin


joshreisner:
a user in #1341 reported an issue deleting the plugin, and i saw that too when trying it on our dev site:

<img width="806" alt="Screenshot 2024-02-07 at 5 01 51 PM" src="https://github.com/code4recovery/12-step-meeting-list/assets/1551689/5a45d7ab-7f08-4a1b-afd4-81dc4e26de59">

kiyote33:
I've played with this a bit.  On a normal pristine installation, it seems to work consistently well.  The above problem happens for me when I've uploaded a new copy of the plugin, activated it, then deactivated it, and finally try to delete the plugin in the normal fashion. I then get the above error.

joshreisner:
this should be fixed now in `3.14.30` (just released). thanks @gkovats !

PR #1345, opened 2024-02-10
[Issue 1342]: Uninstall fix

gkovats:
Addressing issue #1342 :  Apparently when you choose to delete the plugin, the `uninstall.php` is included in isolation, before plugin load. So just adding a require_once to the primary plugin file, to ensure all expected functions are available.

The plugin load happens during ajax request to uninstall. After the request finishes, the files are gone.

### DIFF
--- a/uninstall.php
+++ b/uninstall.php
@@ -5,6 +5,8 @@ if (!defined('WP_UNINSTALL_PLUGIN')) {
     exit();
 }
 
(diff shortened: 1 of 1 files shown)


joshreisner (approved):

joshreisner:
can confirm this resolves the issue, thanks! 🎉

PR #1346, opened 2024-02-11
Block theme template improvements

joshreisner:
fixes TSML UI appearance to work with block themes

makes block theme header CSS output properly

TSML UI before
tsml-ui-before

TSML UI after
tsml-ui-after

Legacy before
legacy-before

Legacy after
legacy-after

### DIFF
--- a/templates/header.php
+++ b/templates/header.php
@@ -4,8 +4,13 @@
 <head>
     <meta charset="<?php bloginfo('charset'); ?>" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <?php wp_head(); ?>
+    <?php
+    // must be before wp_head() to render CSS
+    $header = do_blocks('<!-- wp:template-part {"slug":"header"} /-->');
+    wp_head();
+    ?>
 </head>
 
 <body <?php body_class(); ?>>
-    <?php block_header_area();
+    <?php wp_body_open(); ?>
+    <?php echo $header; ?>
--- a/templates/header.php
+++ b/templates/header.php
@@ -13,4 +13,5 @@
 
 <body <?php body_class(); ?>>
     <?php wp_body_open(); ?>
-    <?php echo $header; ?>
+    <div class="wp-site-blocks">
+        <?php echo $header; ?>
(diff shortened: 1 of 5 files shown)


templates/header.php
```diff
@@ -4,8 +4,13 @@
...
+    // must be before wp_head() to render CSS
+    $header = do_blocks('<!-- wp:template-part {"slug":"header"} /-->');
+    wp_head();
+    ?>
 </head>
 
 <body <?php body_class(); ?>>
-    <?php block_header_area();
\ No newline at end of file
+    <?php wp_body_open(); ?>
```
gkovats, line 15: According to the wiki: https://developer.wordpress.org/reference/functions/wp_body_open/
This was introduced in WP 5.2.0, and our readme says we support from WP 3+ (which man, lot changed since then). Would _maybe_ check before executing:
```
if (function_exists('wp_body_open')) {
     wp_body_open(); 
 }
``` 
Of course, `block_header_area()` wasn't supported until 5.9, so maybe we're out of the woods on this one.

gkovats (approved):
Threw one comment about possibly checking the function. I saw that we're doing that in tsml_header(). But probably a nothingburger. Just loaded a block theme (new stuff to me), and meetings look fine with / without it.

kiyote33 (approved):
No site configuration necessary now. This is what I get on the headers:
TSML:
image
TSML-UI:
image
FYI: The above tests were on Twenty Four. Twenty Two & Three are still a little problematic in getting to look correct.

templates/header.php
joshreisner: yah thanks, good callout. this file only gets invoked if the version is 5.9 above because it checks if `wp_is_block_theme` is a function and true, so i think we're ok

https://github.com/code4recovery/12-step-meeting-list/blob/main/includes/functions.php#L2386

joshreisner:
ah, thanks. looks like `<div class="wp-site-blocks">` needs to surround the header and footer in 2023

updated

2022:


2023:


2024:

kiyote33:
I tested TSML & TSML_UI against patch 31 with 4 block themes and have given them the following ease-of-use ratings:
24: 5 out of 5
23: 4 out of 5
22: 3 out of 5
Spectra 1: 5 out of 5
Think that's pretty good. The first two block themes (22 + 23) from WordPress seem to take more tweaking.

joshreisner:
ok! well now at least the appearance in consistent between legacy and TSML UI on all the WP themes

if we can think of other improvements we can make them later

PR #1351, opened 2024-02-14
Labels to  tsml_location and  tsml_group post types

seanalansnyder:
Added labels as discussed in issue, slack.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -337,6 +337,22 @@ function tsml_custom_post_types()
     register_post_type(
         'tsml_location',
         [
+            'labels' => [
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -337,6 +337,22 @@ function tsml_custom_post_types()
     register_post_type(
         'tsml_location',
         [
+            'labels' => [
+                'name' => __('Locationss', '12-step-meeting-list'),
```
joshreisner, line 341: got an extra "s" here

seanalansnyder: Derp!  Correcting...

joshreisner (approved):

PR #1356, opened 2024-02-16
fix warnings and allow removing conference url

### LINKED ISSUE #1354: Online meeting URL's can't be deleted
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1353

<sup>Originally posted by **ljread** February 15, 2024</sup>
We just moved to a new web host provider. Don't know if this issue is related.

When trying to remove a URL for an online or hybrid meeting, the URL repopulates after pressing the update button (which sometimes requires pressing twice to get an update notification.) Changing or adding a URL works. All other editing functions appear to work. Deleting the meeting and adding back works. TSML appears to display and function normally.

I had thought to delete and reinstall the TSML plugin but would rather avoid this if I can. Any ideas to fix without going to this extreme would be appreciated.

I will try to logon to the old server and see if the issue occurs there as well. I also see there is a new version of TSML, so I will update and hope that fixes my issue.

Wordpress version: 6.4.3
TSML version: 3.14.30

Thanks, John Read</div>

joshreisner:
fix #1354 bug preventing people from removing conference URLs

i notice that i do have to click the Update button twice, not sure if that's new

### DIFF
--- a/12-step-meeting-list.php
+++ b/12-step-meeting-list.php
@@ -4,7 +4,7 @@
  * Plugin Name: 12 Step Meeting List
  * Plugin URI: https://wordpress.org/plugins/12-step-meeting-list/
  * Description: Manage a list of recovery meetings
- * Version: 3.14.31
+ * Version: 3.14.32
  * Requires PHP: 5.6
  * Author: Code for Recovery
  * Author URI: https://github.com/code4recovery/12-step-meeting-list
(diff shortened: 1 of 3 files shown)


kiyote33 (approved):
Me too! Needed to hit the Update twice for it to work as expected. Wonder if that's new or a relic that we never noticed before. Otherwise, works as expected now.

FYI: I checked version 3.14.28 and  the button needed to be clicked twice for the update to happen...

joshreisner:
great thanks

Issue #1366, opened 2024-02-20
Always check that our options are arrays


joshreisner:
see this comment with text from the error: https://github.com/code4recovery/12-step-meeting-list/discussions/1362#discussioncomment-8535792

PR #1371, opened 2024-02-24
Just ensure the option returns an array

gkovats:
For issue #1366, maybe simplistic, but not being able to be certain the `tsml_log` options is always an array, this is just one more step.

### DIFF
--- a/includes/ajax.php
+++ b/includes/ajax.php
@@ -19,9 +19,14 @@ function tsml_ajax_info()
 
     $theme = wp_get_theme();
 
+    $tsml_log = get_option('tsml_log', []);
+    if (!is_array($tsml_log)) {
+        $tsml_log = array();
+    }
+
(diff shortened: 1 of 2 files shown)


joshreisner (approved):

joshreisner:
this looks like a direct way to address one of the two warnings they're seeing in https://github.com/code4recovery/12-step-meeting-list/discussions/1362#discussioncomment-8535792 - it's be great to do the same thing for `array_key_exists`  on functions.php:658

```
    //check cache
    $addresses = get_option('tsml_addresses', []);
    if (!is_array($addresses)) {
      $addresses = array();
    }
```

joshreisner:
there are acutally a lot of places where we could check that get_options returned an array

joshreisner:
i'll merge this and handle that in a separate PR!

PR #1383, opened 2024-03-09
1337 add a meetings block for wp block themes

anchovie91471:
Renders all TSML shortcodes as Gutenberg blocks.

- TSML UI (both versions)
- Regions list
- Upcoming meetings
- Types list

Currently, this PR generates 2 files in the plugin root when running `npx mix --production` or `npx mix watch` (undefined.js & undefined.asset.php). These are unnecessary, and were added to the .gitignore as a stop gap, but this is not ideal. Hoping to get some assistance on finding a better way to compile assets or to fix the anomalous files being generated.
`


assets/build/blocks/block.json
```diff
@@ -0,0 +1,45 @@
...
+      "type": "string",
+      "default": "tsml_ui"
+    },
+    "count": {
+      "type": "string",
+      "default": "5"
+    },
+    "message": {
+      "type": "string",
+      "default": "There are no upcoming meetings today. Please check back tomorrow."
```
joshreisner, line 20: is this text accurate? this seems like it applies more to the Upcoming Meetings widget?

if this block embeds TSML UI it should probably say something like "TSML UI could not be embedded" - not sure what scenario this wouldn't show…

assets/src/blocks/edit.js
```diff
@@ -0,0 +1,70 @@
...
+				<Panel>
+					<PanelBody title={__('Block Settings', '12-step-meeting-list')} icon="admin-plugins">
+						<SelectControl
+							label="Select Block Type"
+							value={props.attributes.blockType}
+							options={[
+								{label: 'Meetings UI', value: 'tsml_ui'},
+								{label: 'Next Meetings', value: 'tsml_next_meetings'},
+								{label: 'Types List', value: 'tsml_types_list'},
+								{label: 'Regions List', value: 'tsml_regions_list'}
```
joshreisner, line 37: do these options show up anywhere? i don't see them in the block editor but i might be looking in the wrong place?

joshreisner (approved):
left a few minor questions

this seems great, i was able to embed the block

i then struggled mightily trying to get it to go wider in 2024 and couldn't, but i don't think that can be helped by us, right? there's no way to make it easier to go wider?

it would be great if we could make a loom or other short screencast video showing how to make the block wider and link it from the description, but that can be in a future PR

assets/build/blocks/block.json
anchovie91471: This is a config value the user can change in the block settings when there are no upcoming meetings listed. It only shows up in the block config for that block, but needs a starting value.

assets/src/blocks/edit.js
anchovie91471: Yes, if you add the "Meetings" block, and click on the block, that setting should show up as a dropdown.


Screenshot 2024-03-23 at 10 02 40 AM

assets/build/blocks/block.json
anchovie91471: This is where that setting shows up.

Screenshot 2024-03-23 at 10 04 21 AM

joshreisner (changes requested):
oh ok i see your response - i'd like to discuss this a little before merging

anchovie91471:
The align toolbar is where you set the width/alignment. I agree, a screenshot at the very least or a short vid might be helpful. I think people who are comfortable with the block editor might know where to look.

I am going to push one more commit to the PR to set .gitignore and webpack.mix.js back to the original files. It looks like my IDE converted tabs to spaces.

Screenshot 2024-03-23 at 10 06 21 AM


> left a few minor questions
> 
> this seems great, i was able to embed the block
> 
> i then struggled mightily trying to get it to go wider in 2024 and couldn't, but i don't think that can be helped by us, right? there's no way to make it easier to go wider?
> 
> it would be great if we could make a loom or other short screencast video showing how to make the block wider and link it from the description, but that can be in a future PR

anchovie91471:
> oh ok i see your response - i'd like to discuss this a little before merging

I think we should. What I did, eliminates the extra files generated by Mix, but adds a build step so we may want to discuss that in addition to whatever you'd like to discuss.

joshreisner:
maybe i'm trying to add it to the wrong place, but i don't see width options on the block

Screenshot 2024-03-23 at 8 14 32 AM

nor on the columns parent

Screenshot 2024-03-23 at 8 14 45 AM

anchovie91471:
What theme are you using? 2024?

anchovie91471:
I didn't realize I should have added support for alignWide in the block.json. Hopefully the latest commit fixes the issue with the alignment toolbar.

joshreisner:
looks like dependabot's pr made a conflict in `package-lock` probably need to `npm i` to resolve it

i'm not seeing alignment options, but perhaps i'm doing it wrong? maybe we could meet at some point and you could show me how to do it?

also should i still be seeing "select a block type" on the right?
Screenshot 2024-03-30 at 8 41 30 AM

gkovats:
FYI fixed the package-lock.json merge conflict.

kiyote33 (commented):
Hi @anchovie91471, just starting a walk-thru today. Thought I'd give you some feedback before diving into the code. Install went as expected after downloading the PR branch. Initial configuration to drop the Meetings block on the home page of my test site was trivial. I dropped a container block on the page, set it to full width, and dropped the Meetings block into the container.  The default setting for the Meetings block was left at Meetings UI. First look found everything as expected with TSML UI displaying to the full width of the container. 
Moving to the other short codes, I tried them 1 at a time and found the following results:
- **Next Meetings**.  Clicked on link and got a 404 Page Not Found. Reset permalinks and still got the 404. On the Import page found the Data Source corrupted. Dropped the data source and recreated it. Problem solved with the 'Next Meetings' shortcode working as expected.
- **Types List**.  As expected, the list contained the types being used in my test data and clicking on the different types gave me a listing of meetings which had that type checked.
- **Regions List**. Again as expected, the list contained the regions in my test data and clicking on a different region gave me a listing of meetings for that region.

Overall impression, no show stoppers. May need to do more investigation into why the corruption of the data source which seemed to occur between my initial load of the Meetings block and the change of settings to the 'Next Meetings' shortcode. I will do the investigation to find out if this is repeatable and needs to be addressed.

Before I dive into the code, could we get a merge of 3.15.1 into your code so I can test that everything works with our latest TSML? Much appreciate your work on this...

kiyote33 (changes requested):
Not 100% sure the corrupted data source issue reported in an earlier comment is a bug related to this code, but it is repeatable. Here's the steps I took to repeat after 1st verifying I had a pristine environment with 36 test records:
1. Deactivated and removed the TSML plugin from the Installed Plugins folder.
2. After confirmation that the TSML plugin was removed, reinstalled and activated the plugin from the zip file in my Downloads folder. 
3. Checked the Imports page for the Data Source status which showed 36 records available from the test data source JSON feed. Verified that the Meeting List showed the 36 records.
4. Switched the Meetings block from the default Meetings UI to the Next Meetings setting.
5. Returned to my home page listing which had the next 5 meetings available listed.
6. Clicked on the first in the listing and promptly got the 404 again.
7. Verified that there are no meeting or regions available in the Meetings editor.
Seems obvious that the meetings and regions are being blown away on deactivation or removal of the TSML plugin, but that the cache file remains, as well as the Data Source settings in the Options table. This is probably not related to this PR, but I think it does need to be resolved.

anchovie91471:
Wow! Thanks for the thorough review @kiyote33 and thanks @gkovats for stepping in on the merge conflict. I completely missed that.

Yep, I'll get the PR up to date with the latest TSML so we can test against that.

One thing I recall @joshreisner mentioned to me when we discussed this a while back, was splitting the the block into 4 separate embeddable blocks, instead of just one with the dropdown. I can certainly do that if the consensus is that would be more user friendly.

kiyote33 (commented):
I like Josh's suggestion to have a separate block for each shortcode. I agree it would be more user friendly. Can we get a block for TSML too? Or do we need a shortcode for it first?

anchovie91471:
> I like Josh's suggestion to have a separate block for each shortcode. I agree it would be more user friendly. Can we get a block for TSML too? Or do we need a shortcode for it first?

Do you mean the actual meeting list? (TSMI UI or Legacy). If so, there should be a choice in the dropdown for Meetings UI, which will render whichever is chosen in the settings for TSML.

kiyote33 (commented):
Yep. Currently we only get TSML UI. It doesn't reference the TSML display setting at all. So there's no way to get the Legacy UI display if it's set to TSML UI.

joshreisner:
since we have a goal of deprecating legacy, my vote would be to have the block be TSML UI only

anchovie91471:
OK. I've merge latest main into the branch @kiyote33

kiyote33 (approved):
Works well in both classic and block themes.

joshreisner (approved):
awesome! thanks!

Issue #1399, opened 2024-03-24
[Bug]: Deprecated


AquaZebra:
### Contact Details

mark@aquazebra.com

### Website With Issue

https://aainthedesert.org/meetings/choice-is-yours-meditation-meeting/

### What happened?

Page without the bug.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Relevant log output/errors

```shell
Deprecated: htmlentities(): Passing null to parameter #1 ($string) of type string is deprecated in /home4/kmpmitmy/public_html/wp-content/plugins/12-step-meeting-list/includes/functions.php on line 1048
```

gkovats:
Looking now, that's certainly a helpful PHP error.

gkovats:
@AquaZebra for what it's worth, this is more a PHP warning we'll fix, but you may want to disable display of errors on your live site:
https://stackoverflow.com/questions/1987579/remove-warning-messages-in-php

That's just cosmetic though, the warnings will still exist, and depending on your host provider this may not be realistic.

anchovie91471:
You may also want to add:

define('WP_DEBUG', false);
define('WP_DEBUG_DISPLAY', false);

to your wp-config.php to remove warnings and errors from showing up on your site. This is a wordpress security best practice.

AquaZebra:
Thank you.

I added this to the wp-config.php file and the errors are no longer showing on the site.
The code was inserted before:
/* That's all, stop editing! Happy blogging. */ in the wp-config.php <https://developer.wordpress.org/advanced-administration/wordpress/wp-config/> file.

// Enable WP_DEBUG mode
define( 'WP_DEBUG', true );
// Enable Debug logging to the /wp-content/debug.log file
define( 'WP_DEBUG_LOG', true );
// Disable display of errors and warnings
define( 'WP_DEBUG_DISPLAY', false );
@ini_set( 'display_errors', 0 );


In Gratitude,

Mark E. Anderson
760-880-5174
***@***.***
https://aquazebra.com


￼

We earn our stripes every day
Helping authors turn their manuscript into a book.
Helping businesses capitalize their online presence by building websites that work.

On Mar 25, 2024, at 7:11 AM, Anthony Baggett ***@***.***> wrote:


You may also want to add:

define('WP_DEBUG', false);
define('WP_DEBUG_DISPLAY', false);

[...]

PR #1400, opened 2024-03-25
Catching random post meta

gkovats:
Issue #1399, you really can't trust what shows up on post_meta. In the user's case, it was something that resolved to a non-string, and then PHP warning'd in the `htmlentities()` step. So adding a few more lines to handle oddball data in post_meta.

Also saw a PHP warning on save, so added a check there. Sorry for more lines of code, functions.php is getting kinda beefy.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1044,11 +1044,18 @@ function tsml_get_meeting($meeting_id = false)
     if (empty($meeting->approximate)) $meeting->approximate = 'no';
 
     //escape meeting values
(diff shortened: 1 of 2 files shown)


joshreisner (approved):
thanks! i cant seem to reproduce this locally - do you have any hypothesis how those null values got into the postmeta?

i see that we get all postmeta for the meeting and run htmlentities on it, so perhaps something outside of TSML set its own metadata?

```php
$custom = get_post_meta($meeting->ID);
```

PR #1404, opened 2024-03-29
[Issue 1402] gettext / POT update

gkovats:
Bumped POT filed from 497 strings to 637 strings. Ran:
```
wp i18n make-pot . ./languages/12-step-meeting-list.pot --exclude=assets/
```
from plugin root to regenerate POT file. This should help folks eager to translate get a little more coverage.


joshreisner (approved):
thanks! tbh i thought string extraction happened automatically on the server

i think there's more work to do here, but this is a great start

future actions could be:
* move all dynamic variables out of strings
* shorten or remove strings wherever possible
* clean up obsolete features like Reopen for In-Person Attendees
* move longer documentation (for example the JSON feeds stuff) to a page/post on our site
* stop using abbreviated terms like "eg"

PR #1422, opened 2024-05-14
Add Timezone Field

joshreisner:
adds:
* optional timezone dropdown to meeting edit screen
* timezone key to JSON (when present)
* timezone column in export CSV
* timezone import

here it is on the edit screen (a future PR could prettify the dropdown to group it by continent and remove underscores):
Screenshot 2024-05-17 at 7 29 25 AM

only TSML UI is affected. in this example the site's timezone is California, and the meeting is America/New_York, and the time is 7pm, so the meeting shows up at 4pm in the list and on the detail screen:

Screenshot 2024-05-17 at 7 29 23 AM

### DIFF
--- a/includes/admin_meeting.php
+++ b/includes/admin_meeting.php
@@ -87,9 +87,7 @@ add_action('admin_init', function () {
                     class="checkboxes<?php if (!empty($tsml_types_in_use) && count($tsml_types_in_use) !== count($tsml_programs[$tsml_program]['types'])) { ?> has_more<?php } ?>">
                     <?php
                     foreach ($tsml_programs[$tsml_program]['types'] as $key => $type) {
-                        if ($key == 'ONL' || $key == 'TC') {
-                            continue;
-                        } //hide "Online Meeting" since it's not manually settable, neither is location Temporarily Closed
+                        if ($key == 'ONL' || $key == 'TC') continue; //hide "Online Meeting" since it's not manually settable, neither is location Temporarily Closed
                         ?>
                         <label <?php if (!empty($tsml_types_in_use) && !in_array($key, $tsml_types_in_use)) {
                             echo ' class="not_in_use"';
@@ -264,6 +262,18 @@ add_action('admin_init', function () {
         <?php } ?>
 
         <div class="meta_form_row">
+            <label>
+                <?php _e('Map', '12-step-meeting-list') ?>
+            </label>
+            <div id="map">
+                <?php if (empty($tsml_mapbox_key) && empty($tsml_google_maps_key)) { ?>
+                    <p>Enable maps on the <a href="<?php echo admin_url('edit.php?post_type=tsml_meeting&page=import') ?>">Import &
+                            Settings</a> page.</p>
+                <?php } ?>
+            </div>
+        </div>
+
+        <div class="meta_form_row">
             <label for="timezone">
                 <?php _e('Timezone', '12-step-meeting-list') ?>
             </label>
@@ -279,18 +289,6 @@ add_action('admin_init', function () {
                     </option>
                 <?php } ?>
             </select>
-        </div>
-
-        <div class="meta_form_row">
-            <label>
-                <?php _e('Map', '12-step-meeting-list') ?>
-            </label>
-            <div id="map">
-                <?php if (empty($tsml_mapbox_key) && empty($tsml_google_maps_key)) { ?>
-                    <p>Enable maps on the <a href="<?php echo admin_url('edit.php?post_type=tsml_meeting&page=import') ?>">Import &
-                            Settings</a> page.</p>
-                <?php } ?>
-            </div>
         </div>
 
         <?php if (count($meetings) > 1) { ?>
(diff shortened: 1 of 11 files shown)


includes/admin_meeting.php
```diff
@@ -261,6 +261,26 @@ function () {
...
+        echo '<option value="' . htmlspecialchars($timezone) . '"';
+        if ($timezone === $selected_timezone) {
+            echo ' selected';
+        }
+        echo '>' . htmlspecialchars($timezone) . '</option>';
+    }
+				?>
+            </select>
+		</div>
+
```
joshreisner, line 283: if you could fix the indentation here that would be great

includes/admin_meeting.php
```diff
@@ -261,6 +261,26 @@ function () {
...
+			</label>
+
+            <select name="timezone" id="timezone">
+				<?php
+				$selected_timezone = isset($meeting->timezone) ? $meeting->timezone : ''; // Allow null tz for backwards compatibility
+				$timezones = DateTimeZone::listIdentifiers(); 
+				foreach ($timezones as $timezone) {
+        echo '<option value="' . htmlspecialchars($timezone) . '"';
+        if ($timezone === $selected_timezone) {
+            echo ' selected';
```
joshreisner, line 276: to be consistent with other dropdowns, it would be great to use wordpress's `selected` function: https://developer.wordpress.org/reference/functions/selected/

includes/save.php
```diff
@@ -298,6 +296,16 @@
         }
     }
 
+	// save timezone
+	if (!$update || strcmp($old_meeting->timezone, $_POST['timezone']) !==0) {
+		$changes[] = 'timezone'	;
```
joshreisner, line 301: there should be a space before the `0` and no space before the `;` - devsense can help!

includes/save.php
```diff
@@ -298,6 +296,16 @@
...
+		if (empty($_POST['timezone'])) {
+			delete_post_meta($post->ID, 'timezone');
+		} else {
+			update_post_meta($post->ID, 'timezone', $_POST['timezone']);
```
joshreisner, line 305: i think `timezone` needs to be sanitized. on line 47 we have an array of strings for sanitization, if you could add `timezone` to that array that should do it

joshreisner (commented):
looks good so far, thanks! still need to add the timezone key, when present, to the json output

includes/save.php
gobborg: I ran php-cs-fixer, and it caught those and also changed indentation.
LMK what you think when I push changes.

includes/save.php
gobborg: I thought about that, but because it's from a dropdown and cannot take text input from a user (unlike Time fields), I decided against it. Happy to change it if you still think it should be sanitized.

includes/admin_meeting.php
gobborg: php-cs-fixer caught it and a few others that I didn't touch. I hope we don't have conflicting linters.

includes/save.php
joshreisner: yes, our policy is to sanitize every input that goes to the database to defend against sql injection, if you see any that we're not sanitizing let us know!

in this case it actually might be better to replace

```
		if (empty($_POST['timezone'])) {
```

with 
```
		if (!in_array($_POST['timezone'], DateTimeZone::listIdentifiers())) {
```

that way we can be extra sure we're getting the data we want

includes/admin_meeting.php
kiyote33, line 1: I'd like to see placeholder text when nothing selected. Something like:
 <option value="" disabled selected>Select meeting timezone when different from site default...</option>
 so users will know they don't always need to make a timezone selection. Does that make sense?

joshreisner: we could add a label with some explanatory text - but some things to think about:

* eventually the meeting finder default timezone will be the set on the Meetings > Settings page, and might be empty (which allows the times to be translated into the user's time)
* eventually this will be set automatically by the API based on your location

given that, i'm not sure if the site default guidance will apply for long. for now i'm thinking of this as a MVP for "pro" users…

includes/variables.php
kiyote33, line 1: 'Time Zone' or 'Timezone', that is the question?

kiyote33 (commented):
Nice work George.

includes/variables.php
joshreisner: "Timezone" is the American usage and what we use in the spec

includes/admin_meeting.php
kiyote33: I get the perspective your pointing out, Time will tell if it becomes an issue on the discussion forum. We can always add something down the road if that's the case.

includes/variables.php
kiyote33: Precedence in TSML supports 'Timezone' so it's less impactful if we stay with Timezone.

gobborg:
Thanks @joshreisner for your help. Changes LGTM for an MVP.

joshreisner:
new update adds a setting for meeting finder timezone

Screenshot 2024-05-18 at 9 01 18 AM

legacy view sees this:

Screenshot 2024-05-18 at 9 01 34 AM

gobborg (approved):

Issue #1436, opened 2024-06-01
[Bug]: Filter TSML_Custom_Fields from the public feed


kiyote33:
### Contact Details

waynep@code4recovery.org

### Website With Issue

https://code4recovery.org/

### What happened?

Custom fields are appearing in the generated meeting feed.

### What browsers are you seeing the problem on?

Firefox

### Relevant log output/errors

```shell
https://code4recovery.org/wp-admin/admin-ajax.php?action=meetings
```

kiyote33:
Fixed in PR #1437.

PR #1437, opened 2024-06-01
Remove custom fields from meetings feed

kiyote33:
Change logic to only include user defined custom fields when doing a full export. This resolves issue #1436.


joshreisner (approved):
nice 🏆

PR #1445, opened 2024-06-12
Check if mb function exists

### LINKED ISSUE #1444: Check if mb_detect_encoding exists before using it
… also utf8_encode is deprecated

### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1443

<sup>Originally posted by **Mimi154** June 11, 2024</sup>
Help! I am trying to update our local meetings. I want to delete all existing meetings and upload the new CSV file. I have attached the file, a screenshot of the error, and the error details.

WP Backend  Error

[template New Meeting List 5_30_24.csv](https://github.com/user-attachments/files/15798479/template.New.Meeting.List.5_30_24.csv)

Upload CSV When importing...
_don't delete anything
_delete only the meetings, locations, and groups for the regions present in this CSV
_delete all meetings, locations, groups, districts, and regions

I have tried all three options and still receive the same error. Any feedback would be greatly appreciated.

Thanks!

WordPress version 6.5.3
Active theme: Spacious (version 1.9.6)
Current plugin: 12 Step Meeting List (version 3.14.35)
PHP version 7.4.33

Error Details
=============
An error of type E_ERROR was caused in line 1475 of the file /var/www/ca4la/wp-content/plugins/12-step-meeting-list/includes/functions.php. Error message: Uncaught Error: Call to undefined function mb_detect_encoding() in /var/www/ca4la/wp-content/plugins/12-step-meeting-list/includes/functions.php:1475
Stack trace:
#0 [internal function]: {closure}()
#1 /var/www/ca4la/wp-content/plugins/12-step-meeting-list/includes/functions.php(1478): array_walk_recursive()
#2 /var/www/ca4la/wp-content/plugins/12-step-meeting-list/includes/admin_import.php(98): tsml_import_buffer_set()
#3 /var/www/ca4la/wp-includes/class-wp-hook.php(324): tsml_import_page()
#4 /var/www/ca4la/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#5 /var/www/ca4la/wp-includes/plugin.php(517): WP_Hook->do_action()
#6 /var/www/ca4la/wp-admin/admin.php(259): do_action()
#7 /var/www/ca4la/wp-admin/edit.php(10): require_once('/var/www/ca4la/...')
#8 {main}
  thrown</div>

joshreisner:
closes #1444 

for #1443

### DIFF
--- a/12-step-meeting-list.php
+++ b/12-step-meeting-list.php
@@ -4,7 +4,7 @@
  * Plugin Name: 12 Step Meeting List
  * Plugin URI: https://wordpress.org/plugins/12-step-meeting-list/
  * Description: Manage a list of recovery meetings
- * Version: 3.14.35
+ * Version: 3.14.36
  * Requires PHP: 5.6
  * Author: Code for Recovery
  * Author URI: https://github.com/code4recovery/12-step-meeting-list
@@ -18,7 +18,7 @@ define('TSML_MEETING_GUIDE_APP_NOTIFY', 'appsupport@aa.org');
 
 define('TSML_PATH', plugin_dir_path(__FILE__));
 
-define('TSML_VERSION', '3.14.35');
+define('TSML_VERSION', '3.14.36');
 
 define('TSML_MEETINGS_PERMISSION', 'edit_posts');
 
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2081,8 +2081,11 @@ function tsml_sanitize_data_sort($string)
     }
 
     # Unicode-aware lowercase of characters in string
-    if (function_exists('mb_strtolower'))
-    return mb_strtolower($t);
+    if (function_exists('mb_strtolower')) {
+        return mb_strtolower($t);
+    } else {
+        return strtolower($t);
+    }
 }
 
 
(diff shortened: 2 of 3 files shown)


kiyote33 (approved):
Tested against uploaded test file (template.New.Meeting.List.5_30_24.csv). No issues with either this or the 3.14.35 release.
PS... Tested up to WordPress 6.5.4, if we want to update that also.

joshreisner:
thanks! i don't think we need to bump the tested up to for patch versions, as long as we get the major and minor we should be good. (also our [wp diretory listing](https://wordpress.org/plugins/12-step-meeting-list/) somehow already says `Tested up to 6.5.4`)

in terms of the test file, i think this only kicks in with a non-standard encoding. could you maybe try creating a new csv from an excel file on windows?

kiyote33:
> thanks! i don't think we need to bump the tested up to for patch versions, as long as we get the major and minor we should be good. (also our [wp diretory listing](https://wordpress.org/plugins/12-step-meeting-list/) somehow already says `Tested up to 6.5.4`)
> 
> in terms of the test file, i think this only kicks in with a non-standard encoding. could you maybe try creating a new csv from an excel file on windows?
I'm not able to repro the issue at all in the previous release (35). I've tried importing the template file into a different encoding and while everything looks OK, I get an error that "Address or City required.".

PR #1446, opened 2024-06-17
Add EDA as program type

gobborg:
Add EDA (eating disorders anonymous) as a supported program type with associated meeting types


gobborg:
I messed up git and mixed up 2 branches.

PR #1447, opened 2024-06-17
Add EDA as program type with supported meeting types

gobborg:


joshreisner (commented):
it would be better to remove the word "meeting" from the types, because it will repeat the word in the tsml ui page title, e.g. Wednesday Outdoor Meeting Meetings"

id recommend "step study" and "topic discussion"

gobborg:
Thanks for that insight and catch @joshreisner . I updated.

joshreisner (approved):

PR #1454, opened 2024-06-21
Make warning persistent if user has a Google Maps key and is using TSML UI

gobborg:
image

### DIFF
--- a/includes/admin_settings.php
+++ b/includes/admin_settings.php
@@ -207,15 +207,8 @@ if (!function_exists('tsml_settings_page')) {
         <!-- Admin page content should all be inside .wrap -->
         <div class="wrap">
 
-            <h1></h1> <!-- Set alerts here -->
-			<?php if (empty($tsml_mapbox_key) && ($tsml_user_interface == 'tsml_ui')) { ?>
-				<div class="notice notice-warning inline">
-					<p>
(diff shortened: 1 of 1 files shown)


joshreisner:
thanks for picking this up! i think it'd be great to simplify this actually and just make this a paragraph that's always there in the maps box. i kind of think these alerts are a nuisance, and feel spammy to me. wdyt?

Screenshot 2024-06-21 at 3 50 05 PM

joshreisner:
sorry for not flagging that sooner! seeing the other alerts in the screeshot kind of clarified it for me a little

gobborg:
I'm fine with moving it to the map API keys box, but do you still want the display to be conditional on
`if (empty($tsml_mapbox_key) && ($tsml_user_interface == 'tsml_ui'))`
or always part of the paragraph?

joshreisner:
i say let's keep it simple and always show it

gobborg:
Before I make changes, let's proof the paragraph:

"Display of maps requires an authorization key from [Mapbox](https://www.mapbox.com/) or [Google](https://console.cloud.google.com/home/). Please note that TSML UI only supports Mapbox."

joshreisner:
i think it's clearer as two paragraphs, per the screenshot

gobborg:
I should get my eyes checked. Thanks for your patience.

joshreisner (approved):
thanks! it would be cool to fix the indentation if possible. but not a blocker. otherwise it'll get fixed by the next commit on this page

PR #1455, opened 2024-06-23
Update change notification email fields to include all tsml_export_columns.

gobborg:
In the existing code, timezone and conference information are not included in the change notification email fields.
image

With this patch, timezone and conference information are included in the change notification email fields. Using the [$tsml_export_columns](https://github.com/code4recovery/12-step-meeting-list/blob/e613a9230580dfb6b7493bef5051f116f69e165d/includes/variables.php#L103) variable future proofs it for further updates.
image


includes/save.php
```diff
@@ -451,11 +451,8 @@
...
             $message .= sprintf(__('This is to notify you that %s created a <a href="%s">new meeting</a> on the %s site.', '12-step-meeting-list'), $user->display_name, get_permalink($post->ID), get_bloginfo('name'));
         }
-        $message .= '</p><table style="font:14px arial;width:100%;border-collapse:collapse;padding:0;">';
-        $fields = array_merge(
-            ['name', 'day', 'time', 'end_time', 'types', 'notes', 'location', 'formatted_address', 'region', 'location_notes', 'group', 'district', 'group_notes'],
-            array_keys($tsml_contact_fields)
-        );
+        $message .= '</p><table style="font:14px arial;width:100%;border-collapse:collapse;parray_keys($tsml_export_columns)adding:0;">';
+        $fields = array_keys($tsml_export_columns);
         foreach ($fields as $field) {
```
joshreisner, line 456: i think this could be `foreach ($tsml_export_columns as $field => $field_name) {`. then you wouldn't need the changes above

joshreisner: also you could delete the `$field_name = __(ucwords(str_replace('_', ' ', $field)), '12-step-meeting-list');` line below

gobborg: Refactoring lines 455-456 is fine, but L26 still needs $tsml_export_column else the email notif doesn't show any fields: 
image

joshreisner: yah i just meant the change on line 454 - something weird got into the style

includes/save.php
```diff
@@ -445,86 +445,82 @@
...
-            } elseif (!empty($old)) {
-                $message .= '<tr style="border:1px solid #999;background-color:#eee;"><td style="width:150px;padding:5px">' . $field_name . '</td><td style="padding:5px">' . $old . '</td></tr>';
-            }
-        }
-        $message .= '</table>';
-        $subject = $update ? __('Meeting Change Notification', '12-step-meeting-list') : __('New Meeting Notification', '12-step-meeting-list');
-        $subject .= ': ' . sanitize_text_field($_POST['post_title']);
-        tsml_email($tsml_notification_addresses, $subject, $message);
-    }
+		if ($_POST['post_status'] !== 'draft') {
```
joshreisner, line 448: could we do `($_POST['post_status'] === 'publish')` and include it with the if above? that way the diff should be smaller

joshreisner: yep!

gobborg: more parens also didn't do it ... `if ((count($tsml_notification_addresses)) && (count($changes)) && ($_POST['post_status'] === 'publish')) {`

When 3 conditions are on the same line, I'm getting emails for drafts.

joshreisner: i'd try `die($_POST['post_status'])` to see what's going on…

joshreisner:
so this is working great except when a field goes from empty to filled, it sees the `$old` value as the new value

Screenshot 2024-11-01 at 7 09 28 PM

i think it has to do with timing and might require a bigger refactor

joshreisner:
going to close for now but feel free to reopen when it's closer to ready

PR #1462, opened 2024-06-29
3.14.36 release prep

joshreisner:
more detail about changes in 3.14.36
EDA language tweaks

i let PSR-12 do its thing with whitespace and if-nesting


gobborg (approved):
LGTM

PR #1497, opened 2024-08-24
Reformat and add Change Detection Email tweaks

kiyote33:
title: Data Source Changes Detected
text: Please sign in to your website and refresh the [data_source_name] Data Source feed on the Import & Export page.
table: two columns, no heading
footer link text: Refresh [data_source_name] Data Source
time: 12 hour format
notes: 1st word should be New, Changed, or Removed depending on the change mode. All changed fields use the $tsml_export_columns value instead of key.
group name: Remove
Should look something like this:
image

### DIFF
--- a/includes/admin_import.php
+++ b/includes/admin_import.php
@@ -213,8 +213,7 @@ if (!function_exists('tsml_import_page')) {
                     tsml_delete_orphans();
 
                     if (count($change_log) && $tsml_debug) {
-                        $message = __('<h2>' . $data_source_name . __(' Feed Update Report', '12-step-meeting-list') . '</h2>');
-                        $message .= tsml_build_import_change_report($data_source_name, $change_log);
+                        $message = tsml_build_import_change_report($data_source_name, $change_log);
                         tsml_alert($message, 'info');
                     }
                     
@@ -354,11 +353,7 @@ if (!function_exists('tsml_import_page')) {
                                 <?php foreach ($tsml_data_sources as $feed => $properties) { ?>
                                     <tr data-source="<?php echo $feed ?>">
                                         <td class="small ">
-                                            <?php
-                                            $refresh_form_name = 'frm_' . tsml_make_refresh_attr_name(@$properties['name'], 'frm');
-                                            ?>
-
-                                            <form name="<?php echo $refresh_form_name ?>" method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
+                                            <form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
                                                 <?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
                                                 <input type="hidden" name="tsml_add_data_source" value="<?php echo $feed ?>">
                                                 <input type="hidden" name="tsml_add_data_source_name"
--- a/includes/admin_import.php
+++ b/includes/admin_import.php
@@ -213,7 +213,8 @@ if (!function_exists('tsml_import_page')) {
                     tsml_delete_orphans();
 
                     if (count($change_log) && $tsml_debug) {
-                        $message = tsml_build_import_change_report($data_source_name, $change_log);
+                        $message = __('<h2>' . $data_source_name . __(' Feed Update Report', '12-step-meeting-list') . '</h2>');
+                        $message .= tsml_build_import_change_report($data_source_name, $change_log);
                         tsml_alert($message, 'info');
                     }
                     
@@ -353,7 +354,11 @@ if (!function_exists('tsml_import_page')) {
                                 <?php foreach ($tsml_data_sources as $feed => $properties) { ?>
                                     <tr data-source="<?php echo $feed ?>">
                                         <td class="small ">
-                                            <form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
+                                            <?php
+                                            $refresh_form_name = 'frm_' . tsml_make_refresh_attr_name(@$properties['name'], 'frm');
+                                            ?>
+
+                                            <form name="<?php echo $refresh_form_name ?>" method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
                                                 <?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
                                                 <input type="hidden" name="tsml_add_data_source" value="<?php echo $feed ?>">
                                                 <input type="hidden" name="tsml_add_data_source_name"
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2283,7 +2283,7 @@ if (!function_exists('tsml_scan_data_source')) {
                     $message = sprintf(__("<p>Please sign in to your website and refresh the <b> %s </b> feed on the Import & Export page.</p><br>", '12-step-meeting-list'), $data_source_name);
                     $message .= tsml_build_import_change_report($data_source_name, $change_log);
                     $import_page_url = admin_url('/edit.php?post_type=tsml_meeting&page=import');
-                    $button_text = __(' Go to Import & Export page', '12-step-meeting-list');
+                    $button_text = __('Go to Import & Export page', '12-step-meeting-list');
                     $message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;border-radius: 3px; color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>$button_text</a>";
 
                     // send Changes Detected email
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -118,11 +118,7 @@ function tsml_build_import_change_report($data_source_name, $change_log, $embed_
 {
     global $tsml_days;
     if (!$embed_in_email) {
-        $hdr_meeting_day_time = __('Meeting', '12-step-meeting-list');
-        $hdr_slug = __('Slug', '12-step-meeting-list');
-        $hdr_notes = __('Notes', '12-step-meeting-list');
         $message = "<table style=\"margin-bottom:10px;text-align:left;border:1px solid #dddddd;padding: 8px;\" cellspacing=\"5\">";
-        $message .= "<tr><th>$hdr_meeting_day_time</th><th>$hdr_slug</th><th>$hdr_notes</th></tr>";
     } else {
         $message = "<table style=\"width:100%;margin-bottom:10px;text-align:left;border:1px solid #dddddd;padding: 8px;\" cellspacing=\"5\">";
     }
@@ -1569,7 +1565,7 @@ function tsml_sanitize_import_meetings($meetings, $data_source_url = null, $data
             if (isset($meetings[$i][$field])) {
                 $meeting_slug = sanitize_title($meetings[$i][$field]);
             }
-            if ($meeting_slug) { 
+            if ($meeting_slug) {
                 break;
             }
         }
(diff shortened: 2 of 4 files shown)


includes/admin_import.php
```diff
@@ -213,7 +213,8 @@ function tsml_import_page()
                     tsml_delete_orphans();
 
                     if (count($change_log) && $tsml_debug) {
-                        $message = tsml_build_import_change_report($data_source_name, $change_log);
+                        $message = __('<h2>' . $data_source_name . __(' Feed Update Report', '12-step-meeting-list') . '</h2>');
+                        $message .= tsml_build_import_change_report($data_source_name, $change_log);
```
joshreisner, line 217: do we need this change? it just adds a title for the table in debug mode - feel like it'll be obvious what this is because the user just clicked refresh

includes/admin_import.php
```diff
@@ -353,7 +354,11 @@ function tsml_import_page()
                                 <?php foreach ($tsml_data_sources as $feed => $properties) { ?>
                                     <tr data-source="<?php echo $feed ?>">
                                         <td class="small ">
-                                            <form method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
+                                            <?php
+                                            $refresh_form_name = 'frm_' . tsml_make_refresh_attr_name(@$properties['name'], 'frm');
+                                            ?>
+
+                                            <form name="<?php echo $refresh_form_name ?>" method="post" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
```
joshreisner, line 361: do we need this form name? what does it do?

includes/functions.php
```diff
@@ -114,33 +114,35 @@ function tsml_bounds()
...
+    if (!$embed_in_email) {
+        $hdr_meeting_day_time = __('Meeting', '12-step-meeting-list');
+        $hdr_slug = __('Slug', '12-step-meeting-list');
+        $hdr_notes = __('Notes', '12-step-meeting-list');
+        $message = "<table style=\"margin-bottom:10px;text-align:left;border:1px solid #dddddd;padding: 8px;\" cellspacing=\"5\">";
+        $message .= "<tr><th>$hdr_meeting_day_time</th><th>$hdr_slug</th><th>$hdr_notes</th></tr>";
+    } else {
+        $message = "<table style=\"width:100%;margin-bottom:10px;text-align:left;border:1px solid #dddddd;padding: 8px;\" cellspacing=\"5\">";
+    }
+
```
joshreisner, line 129: could we make the appearance the same between the email and when showing it in debug mode? want to try to keep logic to a minimum

includes/functions.php
```diff
@@ -2286,22 +2296,15 @@ function tsml_scan_data_source($data_source_url)
...
-                    $message .= "parent_region: $parent_region <br>";
-                    $message .= "database record count: $data_source_count_meetings <br>";
-                    $feedCount = count($meetings);
-                    $message .= "data source feed count: $feedCount<br>";
-                    $message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
-                    $message .= "<br><b><u>Detected Difference</b></u><br>";
-                    $message .= tsml_build_import_change_report($data_source_name, $change_log);
-                    $import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');
-                    $message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Settings page</a>";
+                    $message = "<br>Please sign in to your website and refresh the <b>$data_source_name</b> feed on the Import & Export page.<br><br>";
```
joshreisner, line 2299: could this be a `<p>`?

includes/functions.php
```diff
@@ -2286,22 +2296,15 @@ function tsml_scan_data_source($data_source_url)
...
+                    $message .= tsml_build_import_change_report($data_source_name, $change_log, true);
+                    $refresh_form_name = tsml_make_refresh_attr_name($data_source_name, 'frm');
```
joshreisner, line 2301: is this variable used?

includes/functions.php
```diff
@@ -2286,22 +2296,15 @@ function tsml_scan_data_source($data_source_url)
...
+                    $import_page_url = admin_url('/edit.php?post_type=tsml_meeting&page=import');
+		    $button_text = __("Refresh $data_source_name Data Source", '12-step-meeting-list');
```
joshreisner, line 2303: can't have variables inside localizable strings - might be easier to take out the variable here? otherwise you want to use `sprintf` eg

```
sprintf(__('Please upload a csv file. Your file ended in .%s.', '12-step-meeting-list'), $extension);
```

readme.txt
```diff
@@ -3,8 +3,8 @@ Contributors: code4recovery
 Donate link: https://code4recovery.org/contribute
 Requires at least: 3.2
 Requires PHP: 5.6
-Tested up to: 6.6
-Stable tag: 3.15
+Tested up to: 6.6.1
```
joshreisner, line 6: we don't need to specify patch versions on `Tested up to` - wordpress only complains on minor versions

includes/admin_import.php
kiyote33: I  like it because sometimes when there's lots of Data Sources, you might walk away for lunch or something and it runs while you're gone. Nice to know which report it is if you have lots of different data sources registered.

includes/admin_import.php
kiyote33: Part of the auto-refresh that we decided to wait on till this week.

includes/functions.php
kiyote33: debug_mode is for the troubleshooter, whereas the email let's the local admin know what's going on. Two different but similar purposes where a common routine work's for both.

kiyote33: Need the slug for trouble-shooting....

includes/functions.php
kiyote33: Yep

includes/functions.php
kiyote33: Not at this time. Part of the auto-refresh effort.

includes/admin_import.php
joshreisner: ah this is still work-in-progress? it would make sense to wait on this until the functionality is ready especially since we just rolled out such a big change

includes/functions.php
kiyote33: Why not?

kiyote33: Ok

includes/functions.php
```diff
@@ -2546,6 +2550,7 @@ function tsml_get_changed_import_meetings($feed_meetings, $data_source_url, $dat
         $change_log[] = array(
             'action'      => 'remove',
             'description' => __('Removed / Missing', '12-step-meeting-list'),
+            'notes' => __('Removed ', '12-step-meeting-list'),
```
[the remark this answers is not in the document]
kiyote33: No, only on Changed.

includes/admin_import.php
kiyote33: How about a compromise here? Kill the title line on the debug report, but keep the header.

includes/functions.php
kiyote33: $refresh_form_name variable removed .

includes/functions.php
kiyote33: Thanks. I've added that to my toolkit.

includes/functions.php
kiyote33: Reminder that the default setting for $tsml_debug is false, so normal situation is just the progress bar will be visible during an Update operation.

readme.txt
kiyote33: .1 removed but the following added on line 304:
= 3.15.1 =
* Reformat 'Change Detection Email' for better legability. [more info](https://github.com/code4recovery/12-step-meeting-list/issues/1497)

includes/functions.php
kiyote33: @joshreisner, I've removed report title and column headers from the debug report to match up with the email report. Only difference remaining is that the Slug is available in the on-screen Debug Report. Can you live with that?

joshreisner: i dont see why they should be different though - for example in the line above, it looksl ike the only difference is if it's embedded in email it gets 100% width, otherwise width is not specified? seems like not enough of a difference to have separate logic

for slugs, does that help you look up the meeting? could we link the meeting name instead?

the usage of the arrow is a little strange, could we take that out? it implies that there is a progression, whereas i think you're just using it as a separator

includes/functions.php
```diff
@@ -2609,7 +2601,7 @@ function tsml_compare_imported_meeting($local_meeting, $import_meeting)
     $diff_fields = array();
     foreach($compare_fields as $field) {
         if ($normalized_meetings[0][$field] !== $normalized_meetings[1][$field]) {
-            $diff_fields[] = $field;
+            $diff_fields[] = $tsml_export_columns[$field];
```
joshreisner, line 2604: think we need an `array_key_exists` here in case the key isn't set for some reason

```
$diff_fields[] = array_key_exists($field, $tsml_export_columns) ? $tsml_export_columns[$field] : $field;
```

readme.txt
joshreisner: i think we want to keep the stable tag and lose the tested up to…

includes/functions.php
kiyote33: - I didn't like the 100% width on the debug screen; now that I think about it, I don't think the email needs it either.
- Thanks for the link idea, that will be much better.
- I liked the arrow separator but sort of expected you wouldn't. Will remove.

readme.txt
```diff
@@ -301,6 +301,9 @@ Yes, you will need to know the key name of the field. Then include an array in y
 
 == Changelog ==
 
+= 3.15.1 =
+* Reformat 'Change Detection Email' for better legability. [more info](https://github.com/code4recovery/12-step-meeting-list/issues/1497)
```
joshreisner, line 305: typo: legibility

includes/functions.php
kiyote33: All the keys and heading values are explicitly set in the Variables file. We remove some key/values when we create the $compare array to , but I don't see the $tsml_export_columns array getting touched again. So, by my reconning there's no way the key value could not be there. Are you sure about this or am I missing something?

includes/functions.php
```diff
@@ -2286,22 +2280,14 @@ function tsml_scan_data_source($data_source_url)
...
-                    $message .= "data source feed count: $feedCount<br>";
-                    $message .= "Last Refresh: <span style='color:red;'>*</span>" . Date("l F j, Y  h:i a", $data_source_last_import) . '<br>';
-                    $message .= "<br><b><u>Detected Difference</b></u><br>";
+                    $message = sprintf(__("<p>Please sign in to your website and refresh the <b> %s </b> feed on the Import & Export page.</p><br>", '12-step-meeting-list'), $data_source_name);
                     $message .= tsml_build_import_change_report($data_source_name, $change_log);
-                    $import_page_url = admin_url('edit.php?post_type=tsml_meeting&page=import');
-                    $message .= "<a href='" . $import_page_url . "' style=' margin: 0 auto;background-color: #4CAF50;border: none;color: white;padding: 25px 32px;text-align: center;text-decoration: none;display: block;font-size: 18px;'>Go to Import & Settings page</a>";
+                    $import_page_url = admin_url('/edit.php?post_type=tsml_meeting&page=import');
+                    $button_text = __(' Go to Import & Export page', '12-step-meeting-list');
```
joshreisner, line 2286: super minor sorry i didnt see this before could we remove the leading space

joshreisner (approved):

Issue #1500, opened 2024-08-27
Tighten up "Confererence URL" validation


joshreisner:
currently we validate conference URLs only by checking if the domain is in a list of approved domains

we should go further with zoom specifically to verify that the url contains `zoom.us/j/` and then 8 or more numbers in row

we are finding that some sites are entering `zoom.us/join` as a URL, which is causing a bad experience for users

this will cause some customer pain, so we need a clear error message to show up client-side when an invalid URL is entered

joshreisner:
looking at data from meeting guide, there is a lot of variation in the URLs we're seeing. i think we want to apply the following transformations to be helpful:

* change `http://` to `https://`
* strip subdomain if present
* strip URL params that are not `pwd=`
* strip `pwd=` if it's not a valid hash (not 32 characters, or, say, between 28 and 36)
* strip anything after `#`
* lowercase `/J/`

however it's an error if the URL is zoom and:
* has more than 4 slashes
* does not start with `http://` or `https://`
* path does not start with `/j/` (or `/J/`)
* path does not end with a 8-12 character number

joshreisner:
closed by @gkovats in #1512

Issue #1002, opened 2022-12-16
Refactor geocoding


joshreisner:
### Requested Feature/Enhancment

Regardless of which service provides the geocoding response, there is an opportunity for us to improve the way we get geocodes:

* javascript errors should not block saving meeting changes
* when there is a geocoding problem, the user should be alerted (not have to check the console)
* when geocoding is not available, location changes should be saved into a 'pending' state (this could be a step to enabling automated JSON imports)

This will require team discussion and refinement to move forward.

### TSML Version

Latest (default)

### Wordpress Version

Latest (default)

PR #1512, opened 2024-09-11
Issue 1500 conference validation

gkovats:
For issue #1500 :
The current form has 1 variable to track valid state for the whole form based on formatted_date, so tracking a second field meant tracking state per field. Also just built a `$fields` object to get all the field selectors up front. Didn't know how far this would go.

- conference_url field state now tracked separately, and validated on page load, so meetings with currently bad Zoom urls will either throw an error / highlight the field and disable publish, or automatically correct the url format. 
- moved one formatted_address field error state to this new method, "In person meetings must have a specific address"
- kept language for error messages in PHP so they can use gettext translation.

### Testing:
- load `main` branch, and add a poorly formatted zoom url, then switch to this branch and load the meeting. The Zoom url should get validated immediately - if bad, will error and block publishing. If close, will get corrected and message will say "Url was cleaned up.."
- Updated pointers for Open / Closed checks, so make sure you can't click both types
- Test the validation on Formatted address: 
  - When online meeting, a full address should show the message "Full addresses will trigger meeting is closed message"
  - When in person meeting, an approximate address should throw an error "must have a full address for in person meetings" and block publishing


assets/src/admin-meeting.scss
```diff
@@ -98,6 +98,17 @@
 		}
 	}
```
gkovats, line 100: used `<small>` tags to follow input fields for immediate error messaging.

includes/admin_meeting.php
```diff
@@ -235,6 +238,12 @@ function () {
             <input value="<?php tsml_echo($location, 'approximate') ?>" type="hidden" name="approximate" id="approximate">
             <input value="<?php tsml_echo($location, 'latitude') ?>" type="hidden" name="latitude" id="latitude">
             <input value="<?php tsml_echo($location, 'latitude') ?>" type="hidden" name="longitude" id="longitude">
+            <small class="error_message" data-message="1">
+                <?php _e('Error: In person meetings must have a specific address.', '12-step-meeting-list') ?>
+            </small>
+            <small class="warning_message" data-message="2">
+                <?php _e('Warning: Unable to process this address for exact location.', '12-step-meeting-list') ?>
+            </small>
```
gkovats, line 246: I added this error state to capture when we get an error from the geocoding call. I figured it was worth displaying, but you may know reasons why this isn't useful - maybe frequent false positives?

package.json
```diff
@@ -1,5 +1,9 @@
 {
 	"private": true,
+    "scripts": {
+        "start": "npx mix watch",
+        "build": "npx mix --production"
+    },
```
gkovats, line 6: Just added `npm start` as a shortcut for mix watch. I'm lazy.

assets/src/admin.js
```diff
@@ -60,26 +60,86 @@ jQuery(function ($) {
 		//make sure geocoding is finished (basic form validation)
 		var form_valid = true;
```
gkovats, line 62: Sorry, this is where I went nuts. Made an object to track all the fields, then a small jQuery function to track / attach "state" to each field, and an upper formstate check, so `formatted_address` and `conference_url` (and other fields) could have issues independently.

assets/src/admin.js
```diff
@@ -60,26 +60,86 @@ jQuery(function ($) {
...
+        // after field states update, toggle publish button
+        function updateFormState() {
+            var pendingFields = [];
+            Object.keys($fields).forEach(function(field) {
+                var state = $fields[field].data('state');
+                if (-1 < ['error','loading'].indexOf(state)) {
+                    pendingFields.push($fields[field][0])
+                }
+            });
+            form_valid = !pendingFields.length;
```
gkovats, line 130: later, we could use `pendingFields` as an array of DOM elements for hyperlinks, to anchor link users to in some top message. "The following fields need attention: ...."

assets/src/admin.js
```diff
@@ -330,17 +388,44 @@ jQuery(function ($) {
...
-			$('input#formatted_address').change();
+		$fields.in_person.on('change', function () {
+			$fields.formatted_address.trigger('change');
 		});
-		$('input#conference_url').change(function () {
-			$('input#formatted_address').change();
-		});
-		$('input#conference_phone').change(function () {
-			$('input#formatted_address').change();
+
```
gkovats, line 394: Zoom error and cleanup logic

kiyote33 (approved):
Good job George. Like the radio button effect on the open/close. Nice touch. Everything worked as described in your prep notes. One little thing with the Full Address warning. Shouldn't that be an error instead of warning? You can't really proceed until you fix. Other than that, I think it's good to go.

joshreisner:
i'm really looking forward to reviewing this after NAATW, thanks @gkovats !

gkovats:
Just a heads up, if you're ever waiting on me to merge a PR, let me know - I won't unless asked. Not my codebase, still new.

gkovats:
@kiyote33 Thanks, I missed that - it _was_ an error, but I didn't treat it as such. Just updated in the latest commit.
image

assets/src/admin-meeting.scss
joshreisner: 👍 thanks for using the same color of red we use in other places - maybe eventually we can use css vars for it

includes/admin_meeting.php
joshreisner: no this is great - we can now cross this requirement off our [refactor geocoding](https://github.com/code4recovery/12-step-meeting-list/issues/1002) ticket

package.json
joshreisner: sounds good - we can make make this standard in other places too

joshreisner (approved):
works great, and thanks for rebuilding the strings

the only problem i had while testing was my old js and css assets were cached, which meant that i saw the error messages on the screen as black text flush left on the page

i am 99% sure that when this gets released our `TSML_VERSION` bump will bust the cache but it's just something to keep an eye on when this goes out

PR #1513, opened 2024-09-12
Issue 1511 Shortcode to Filter Meetings 

kiyote33:
Modified tsml_ui shortcode to accept passed parameters filter meeting available.
Reordered shortcode functions alphabetically for easier access. Added example text. 
Updated version, readme log and FAQ descriptions for shortcodes.

Example shortcode to limit the listing to closed meetings for women only:
 `[tsml_ui key="type" first_value="women" second_value="closed" third_value=""]`
 image

 Here an example to restrict meeting regions to a few sub regions of San Jose
`[tsml_ui key="region" first_value="campbell" second_value="cupertino" third_value="los-gatos"]`
image

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2168,6 +2168,7 @@ function tsml_to_css_classes($types, $prefix = 'type-')
     }
 
     $types = array_map('strtolower', $types);
+    $types = array_map('trim', $types);
 
     return $prefix . implode(' ' . $prefix, $types);
 }
--- a/includes/shortcodes.php
+++ b/includes/shortcodes.php
@@ -6,113 +6,26 @@ add_shortcode('tsml_location_count', 'tsml_count_locations');
 add_shortcode('tsml_meeting_count', 'tsml_count_meetings');
 add_shortcode('tsml_region_count', 'tsml_count_regions');
 
-//output the Legacy UI meeting finder
-//example:  [tsml_query_string key="type" value="women"]
-function tsml_legacy_ui($atts, $content = null)
-{
-    global $tsml_meeting_attendance_options;
-
-    //Load Shortcode values into array, overriding default values
-    $params = shortcode_atts(array('key' => '', 'value' => '', ), $atts);
-
-    $key = (isset($params['key'])) ? sanitize_text_field($params['key']) : '';
-    $value = (isset($params['value'])) ? sanitize_text_field($params['value']) : '';
-
-    $meetings = tsml_get_meetings([
-        $key => strtoupper($value),
-        'attendance_option' => 'active',
-    ]);
-
-    if (!count($meetings)) {
-        return '<div class="tsml-no-upcoming-meetings">No Meetings Available</div>';
-    }
-
-    $rows = '';
-    foreach ($meetings as $meeting) {
-        $meeting_types = $classes = '';
-        if (!empty($meeting['types'])) {
-            $classes = tsml_to_css_classes($meeting['types']);
-            $meeting_types = ' <small><span class="meeting_types">' . tsml_format_types($meeting['types']) . '</span></small>';
-        }
-
-        if (!empty($meeting['notes'])) {
-            $classes .= ' notes';
-        }
-
-        $meeting_location = '';
-        if (!empty($meeting['location'])) {
-            $meeting_location = $meeting['location'];
-        }
-
-        if ($meeting['attendance_option'] == 'online' || $meeting['attendance_option'] == 'inactive') {
-            $meeting_location = !empty($meeting['group']) ? $meeting['group'] : '';
-        }
-
-        $region = '';
-        if (!empty($meeting['sub_region'])) {
-            $region = $meeting['sub_region'];
-        } elseif (!empty($meeting['region'])) {
-            $region = $meeting['region'];
-        }
-
-        $rows .= '<tr class="meeting ' . $classes . ' attendance-' . $meeting['attendance_option'] . '">
-				<td class="time">' . tsml_format_day_and_time($meeting['day'], $meeting['time']) . '</td>
-				<td class="name"><a href="' . $meeting['url'] . '">' . @$meeting['name'] . '</a>' . $meeting_types . '</td>
-				<td class="location">
-					<div class="location-name">' . $meeting_location . '</div>
-					<div class="attendance-option attendance-' . $meeting['attendance_option'] . '"><small>' . ($meeting['attendance_option'] != 'in_person' ? $tsml_meeting_attendance_options[$meeting['attendance_option']] : '') . '</small></div>
-				</td>
-				<td class="region">' . $region . '</td>
-			</tr>';
-    }
-    return '
-	<style>
-		table.tsml_query_string div.attendance-hybrid small,
-		table.tsml_query_string div.attendance-online small {
-			color: green;
-		}
-
-		table.tsml_query_string div.attendance-inactive small {
-			color: #d40047;
-		}
-	</style>
-	<table class="table table-striped">
-		<thead>
-			<tr style="text-align: left;">
-				<th class="time">' . __('Day & Time', '12-step-meeting-list') . '</th>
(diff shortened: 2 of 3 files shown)


includes/functions.php
```diff
@@ -2168,6 +2168,7 @@ function tsml_to_css_classes($types, $prefix = 'type-')
     }
 
     $types = array_map('strtolower', $types);
+    $types = array_map('trim', $types);
```
gkovats, line 2171: @kiyote33 Hey Wayne I know no one asked, but locally I saw this was spitting out odd classes like:
`type-o type- onl type- en type-onl`
because types often seem to have whitespace around them. Hope this is OK.

gkovats (approved):

gkovats:
Looks good locally. Only thing that threw me was the two shortcodes accept different params:
```
[tsml_legacy_ui key="type" value="w"]
[tsml_ui key="type" value="women"]
```
But nothing major.

includes/shortcodes.php
```diff
@@ -6,8 +6,96 @@
 add_shortcode('tsml_meeting_count', 'tsml_count_meetings');
 add_shortcode('tsml_region_count', 'tsml_count_regions');
 
+//output the Legacy UI meeting finder
+//example:  [tsml_query_string key="type" value="women"]
+function tsml_legacy_ui($atts, $content = null)
```
joshreisner, line 11: i would prefer not to add a shortcode for "legacy ui" since it wasn't requested, and we are trying to move away from legacy ui. could we remove this option?

legacy ui doesn't really lend itself to being a shortcode because it's too complex. - it won't be the full thing with filters and search, etc, and clicking back to meetings will take you to a different page

includes/shortcodes.php
```diff
@@ -106,12 +224,13 @@ function tsml_next_meetings($arguments)
     return '<h3>Types</h3><ul>' . implode($types) . '</ul>';
 });
 
-//output a react meeting finder widget https://github.com/code4recovery/tsml-ui
-function tsml_ui()
+/*output a react meeting finder widget https://github.com/code4recovery/tsml-ui
+ *examples: [tsml_ui], with parameters: [tsml_ui key="type" value="women"] */
```
joshreisner, line 228: could the format for this be `[tsml_ui type="women,closed"]`

also if the comment format could be consistent with the other functions (above the definition, with a double slash in front), thanks

includes/shortcodes.php
```diff
@@ -106,12 +224,13 @@ function tsml_next_meetings($arguments)
...
+function tsml_ui($atts, $content = null)
```
joshreisner, line 229: $content doesn't seem to be used. could `$atts` be `$arguments` for consistency with other shortcodes?

includes/functions.php
kiyote33: No I hadn't seen that. Thx for fixing...

includes/shortcodes.php
```diff
@@ -124,24 +243,51 @@ function tsml_ui()
...
+                    ],
                 ],
-            ],
-            $tsml_ui_config
-        )
-    );
+                $tsml_ui_config
+            )
+        );
+    } else {
```
joshreisner, line 271: seems like a lot of code duplication if just `defaults` changes between these implementations. could you try without the else? i think if $types were an empty string then `explode(',', $types)` might be ok

readme.txt
```diff
@@ -301,6 +303,9 @@ Yes, you will need to know the key name of the field. Then include an array in y
 
 == Changelog ==
 
+= 3.15.1 =
+* Create/modified shortcodes to filter meetings by passed types. [more info]
```
joshreisner, line 307: missing link for more info

should be `3.15.2`

joshreisner (changes requested):

includes/shortcodes.php
kiyote33: Yep. I expected you wouldn't like it but it was just so easy to do. I'll get rid of it.

includes/shortcodes.php
kiyote33: $atts is consistent with all the WordPress examples. How about we change that one instance of $arguments to be consistent with $atts?

includes/shortcodes.php
kiyote33: Absolutely! If testing shows that works I will remove the code duplication.

readme.txt
kiyote33: Thx. Will fix...

includes/shortcodes.php
kiyote33: Just for the record, I don't support moving away from TSML. I think the majority of sites are based on it and we have a long ways to go before TSML UI has all the features which would allow TSML to be depricated...

includes/shortcodes.php
kiyote33: Could not make the suggested format work.  I was able to pass the two types as a single parameter, but unable to get the formatting right to generate the tsml_react_config object. 
Commenting format fixed.

includes/shortcodes.php
kiyote33: Unused $content parm removed.

includes/shortcodes.php
kiyote33: Duplicate wp_localize_script code removed after testing showed it wasn't needed. Only needed to add 'Any Type' as the default value when no value is received in the $atts array.

includes/shortcodes.php
joshreisner: i believe that `$arguments` is better than `$atts` because it's clearer and not colloquial. same with `$parms` - it should be `$params` or, better, `$types`

keeping our variable names clear helps developer cognitive load 🙏

kiyote33: Sorry, but I read somewhere that there was an advantage to using $atts rather than something like $args because there is some behind-the-scene preprocessing that goes on.  I think it had todo with the output validation, but I can't remember the details. I can't find the article this morning, but if I come across it I'll send it on to you.

includes/shortcodes.php
joshreisner: `['women', 'closed']` should filter to meetings that are both women and closed in TSML UI

kiyote33: Done

joshreisner: looks like the pattern you're suggesting is `[tsml_ui key="type" first_value="women" second_value="closed" third_value=""]` but i still really think it should be `[tsml_ui type="women,closed"]`

joshreisner:
```
    //get parameters
    $arguments = shortcode_atts(['type' => ''], $arguments, 'tsml_ui');
    $defaults = [
        'type' => array_map('sanitize_title', explode(',', $arguments['type']))
    ];
```

kiyote33: I think the checked in code is much clearer, and most importanly: it work's...

joshreisner: wow, i'm sorry you feel that way. looks like we're at an impasse then

kiyote33: Yea! I'm sorry too...

gkovats: to be fair, `[tsm_ui key="type" value="women,closed"]` would be the first thing I'd try. The other feels a little obtuse. Feels worth a few more lines of code to follow how the shortcode would be naturally used, just going off other's I work with.

gkovats: BUT, hey, maybe fodder for the next standup?

joshreisner: thanks @gkovats but how would you handle it if they asked for more filters? my approach allows room for more, eg

`[tsml_ui type="lgbtq" weekday="tuesday" region="usa"]`

it seems like a code smell to use two arguments that need to work in combination when one is more direct

gkovats: oh, yeah, when I saw the shortcode figured `key="type"` was a standard. ^ that'd be even more familar, key value pairs. Sorry, my original was "it works" without context. Yeah, then we'd just have accepted shortcode attributes to look.

Issue #1074, opened 2023-02-11
Enforce PayPal username format


joshreisner:
### Requested Feature/Enhancement

https://www.paypal.com/us/smarthelp/article/what-is-paypalme-help432

if the content a user has entered contains an invalid paypal.me username, show a warning, and don't save it

verify that valid paypal usernames form a button in legacy view and go to the right place

joshreisner:
closed by @gkovats in #1524

PR #1524, opened 2024-09-21
Issue #1074 Adding paypal validation

gkovats:
Issue #1074 - just adding a check for PayPal format.

And cause I can't leave well enough alone, added 3 lines for `$fields.[name].clearState()` because it felt more intuitive. It's kinda low on the page.

**Note**: I did trigger this on load, to call attention to bad PayPal usernames as we did for Zoom urls. If this is too aggressive, would just remove the check on load.


gkovats:
Also, whitespace is tabs and spaces. I used spaces since PHP is using that.

assets/src/admin.js
```diff
@@ -425,6 +430,20 @@ jQuery(function ($) {
 			$fields.formatted_address.trigger('change');
 		});
 
+        // validate paypal name
+        $fields.paypal.validate = function() {
+            var value = $fields.paypal.val().trim();
+            // must be letters and numbers, under 20 characters
+            // reference: https://www.paypal.com/us/cshelp/article/what-is-paypalme-help432
```
joshreisner, line 437: thanks for the link - this will be helpful in the future

joshreisner (approved):

PR #1526, opened 2024-09-23
Issue 1498 - Adding entity fields

### LINKED ISSUE #1498: Add Support for Service Entity Fields in TSML
The  following fields are part of the Meeting Guide App spec and support needs to be added to TSML:

## Service Entity Fields
`entity`
Entity is the name of the service entity responsible for the listing. entity info is optional, but entity is required if any of the other entity fields are present.

`entity_email`
Entity Email is a public email address for the service entity responsible for the listing. This should be a single email address.

`entity_feedback_emails`
Entity Feedback Emails is an array of feedback addresses for the service entity responsible for the listing. When using a Google Sheet, separate multiple addresses with a ,.

`entity_location`
Entity Location is a human-readable physical description of the service area of the entity, for example Whatcom County, Washington.

`entity_logo`
Entity Logo is the URL of the logo of the service entity responsible for the listing. It should begin with https://. Ideally the image this points to is a vector-based SVG so it can be scaled to any size. Additionally, the image should be square, and have a transparent background. Finally, colors should be specified using currentColor so that they can adapt to the color mode (light, dark) of the app.

`entity_phone`
Entity Phone is the phone number of the service entity responsible for the listing. Should be in the format +1-123-456-7890 and start with country code for international dialing.

`entity_website_url`
Entity Website URL is the website homepage of the service entity responsible for the listing. This should begin with `https://`.

gkovats:
- Closes #1498

Adds Entity form fields to Settings page:
- 5 fields: entity, entity_email, entity_url, entity_phone, entity_location
- limits all to 100 characters: wanted to limit because this will add weight to every meeting in every feed.
  - *Note*: the added weight is worth bringing up to Meeting Guide, they'll likely notice
- added defaults for entity, entity_url, entity_email, so we start with values from WordPress

Adds Entity fields to export / import
- sanitizer caps strings off at 100 characters
- export supplies the local entity values for meetings without a data_source, otherwise grabs any entity values from post_meta / meeting fields

Adds Entity fields to TSML Legacy
- Added in "Updated" box. I took a crack at it, seemed like it belonged their, but I know we want to be efficient with screen real estate, and it should follow the pattern that TSML UI and Meeting Guide will use. Consider this a draft.

image


includes/admin_settings.php
```diff
@@ -332,7 +346,59 @@ function tsml_settings_page()
                         </form>
                     </div>
```
gkovats, line 348: Check the order of these on the settings screen. I ended up putting this in spot 2, left column, and moved the Feeds settings to top of the middle column. But I don't stare at this screen much, so it's a guess.

includes/functions.php
```diff
@@ -1774,6 +1791,11 @@ function tsml_sanitize_import_meetings($meetings, $data_source_url = null, $data
             }
         }
 
+        //cleanup entity fields
+        foreach($tsml_entity_fields as $field) {
```
gkovats, line 1795: I can defend this, but agree it looks awful.

includes/admin_settings.php
```diff
@@ -332,7 +347,56 @@ function tsml_settings_page()
...
+                        <!-- Entity Form -->
+                        <?php
+                        $tsml_entity = tsml_get_entity();
+                        ?>
+                        <div class="stack compact">
+                            <h2>
+                                <?php _e('Service Entity Information', '12-step-meeting-list') ?>
+                            </h2>
+                            <p>
+                                <?php echo sprintf(__('Enter information for your service entity here to help identity the meeting source when sharing feeds with others.', '12-step-meeting-list'), get_post_type_archive_link('tsml_meeting')) ?>
```
joshreisner, line 360: don't need the `sprintf` here - we'd only need that if you were adding params to the i18n string

includes/admin_settings.php
```diff
@@ -202,6 +202,21 @@ function tsml_settings_page()
...
+            global $tsml_entity_fields;
+            foreach ( $tsml_entity_fields as $field ) {
+                $tsml_entity[$field] = isset($current_tsml_entity[$field]) ? $current_tsml_entity[$field] : '';
+                if (isset($_POST["tsml_$field"])) {
+                    $tsml_entity[$field] = substr(trim(strval($_POST["tsml_$field"])), 0, 100);
+                }
+            }
+            update_option('tsml_entity', $tsml_entity);
+            tsml_cache_rebuild(); //these values affects what's in the cache
+        }
```
joshreisner, line 219: i think we want a `tsml_alert` here to say that the update was successful

templates/single-meetings.php
```diff
@@ -359,8 +359,35 @@ class="btn btn-default btn-block contact-phone">
...
                                 } ?>
                                 <li class="list-group-item list-group-item-updated">
-                                    <?php _e('Updated', '12-step-meeting-list') ?>
-                                    <?php the_modified_date() ?>
+                                    <p class="meeting-updated">
+                                        <?php _e('Updated', '12-step-meeting-list') ?>
+                                        <?php the_modified_date() ?>
+                                    </p>
+                                    <?php
+                                    if (!empty($meeting->{'data_source'})) {
```
joshreisner, line 367: i think we need some intro text here potentially like "This meeting listing is provided by:"

let's show the email address - this is more for administrative purposes

the feedback email form should either be hidden and we add something here, or it should send to the entity's feedback_email addresses if present

minor: could we use the `$meeting->data_source` pattern for this and the other vars?

minor: i would like to steer away from using `<br>` tags - could this be Ps or a UL?

includes/admin_settings.php
```diff
@@ -332,7 +347,56 @@ function tsml_settings_page()
...
+                            <h2>
+                                <?php _e('Service Entity Information', '12-step-meeting-list') ?>
+                            </h2>
+                            <p>
+                                <?php echo sprintf(__('Enter information for your service entity here to help identity the meeting source when sharing feeds with others.', '12-step-meeting-list'), get_post_type_archive_link('tsml_meeting')) ?>
+                            </p>
+
+                            <form method="post" class="stack compact" action="<?php echo $_SERVER['REQUEST_URI'] ?>">
+                                <?php wp_nonce_field($tsml_nonce, 'tsml_nonce', false) ?>
+                                <h3>
```
joshreisner, line 365: could these `h3`s be `label`s instead? the smaller type should look good as well

i would keep it `Entity Name` but change:

`Entity Contact Email` -> `Administrative Contact Email`
`Entity Contact Phone` -> `Public Phone Number`
`Entity Location` -> `Your Service Area`
`Entity Website` -> `Website Address`

includes/variables.php
```diff
@@ -76,6 +76,15 @@
...
 }
 
+//define entity fields (stored in option tsml_entity)
+$tsml_entity_fields = [
+    'entity',
+    'entity_email',
+    'entity_phone',
+    'entity_location',
+    'entity_url',
+];
```
joshreisner, line 86: do we need to add feedback email here? the idea is that they should pass through from the imported service entity

joshreisner (commented):
look great, thanks for jumping on this so quickly! just a few comments

templates/single-meetings.php
gkovats: Adding "This meeting listing is provided by:" only concern it will show up everywhere if we display for local _and_ imported meetings, and it just adds that one more line to what your eye is scanning for
image

Look correct to you? Should it be smaller?

gkovats: > or it should send to the entity's feedback_email addresses if present

Honestly this raises questions I want to talk about, cause it'll be a mess typing it out. Don't feel equipped with enough history on this tool to answer them myself.

includes/admin_settings.php
gkovats: "Your" felt redundant on looking at it, like it could be in all labels. This OK, or you think the "your" belongs?
image

includes/functions.php
```diff
@@ -1774,6 +1790,11 @@ function tsml_sanitize_import_meetings($meetings, $data_source_url = null, $data
...
+            $meetings[$i][$field] = substr(trim(strval(isset($meetings[$i][$field]) ? $meetings[$i][$field] : '')), 0, 100);
```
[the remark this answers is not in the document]
gkovats: Thanks, updated.

templates/single-meetings.php
joshreisner: hm, yeah, i'm not so sure. perhaps less is more?

my reason for wanting to add it is that we want to make it clear what this section is - some users might get confused and think it is meeting information.

the purpose of this section is to be like this section in the app:
Screenshot 2024-09-24 at 3 11 42 PM

if we could replicate the headline style for the other sections for the entity name, and make the url and phone buttons like in this pic:

Screenshot 2024-09-24 at 3 14 05 PM

we might try the `This meeting listing is provided by:` above the headline in a `<small>`?

includes/admin_settings.php
joshreisner: that's good! maybe make them `label`s though, if they're not already

templates/single-meetings.php
joshreisner: here's what it currently looks like in [this PR for TSML UI](https://github.com/code4recovery/tsml-ui/pull/416):

image

not that it needs to match per se but just for context

joshreisner:
one minor thing: i'm seeing PHP warnings on some of the entity form fields when running locally:

Screenshot 2024-10-12 at 8 59 23 AM

joshreisner:
also i think `feedback_emails` should use `array_values` - see example at https://demo.code4recovery.org/wp-admin/admin-ajax.php?action=meetings

Screenshot 2024-10-12 at 9 07 16 AM

joshreisner:
one last piece of feedback, and feel free to push back and/or say we should make this a subsequent PR, is i think if feedback emails are present on a meeting in legacy, we should expose the feedback form and send to those emails.

ie here:
```
                        if (!empty($tsml_feedback_addresses)) { ?>
```
we might say 
```
                        if (!empty($tsml_feedback_addresses) || !empty($meeting->feedback_emails) { ?>
```

gkovats:
@joshreisner  FYI
- added `array_values()` to places where I'm casting to array to avoid keys
- fixed the admin settings form to ensure those array fields are present before displaying
- updated form button to use already existing button style class `btn btn-default btn-block` and removed my new style
- fixed the logic for displaying the change request form - it was entirely an oversight on my part

joshreisner:
looks awesome! only detail i noticed is the link icon is a little big - looks like it just needs width and height props

if you don't mind i'll just go ahead and merge this add them in the follow-on PR where i set the new version

Screenshot 2024-10-13 at 7 41 02 PM

joshreisner (approved):

PR #1528, opened 2024-09-25
Allow invalid zoom urls for now

joshreisner:
in talking with GSO yesterday, it seemed the more moderate course for now would be to allow invalid zoom URLs for now, and just display the warning

leaving the formatting alone for the moment - at some point this file should be normalized with prettier


gkovats (approved):
Womp womp. I get their concern.

PR #1539, opened 2024-10-13
make tsml ui load async

joshreisner:
not sure why we never made it load async before 

Screenshot 2024-10-13 at 8 24 11 AM

i considered stopping it from appending the wordpress version here, but perhaps that is useful info for us to have?


gkovats (approved):
worked on my end

PR #1551, opened 2024-10-26
add Google Sheets directly

### LINKED ISSUE #1348: Import Google Sheets by URL
Currently it's possible to import a Google Sheet directly to TSML by using the Google Sheets API, more info here: https://github.com/code4recovery/12-step-meeting-list/pull/760#issue-1204556731

It would be far simpler for users to be able to paste a sheet URL, but we want to avoid adding yet another Google API key to our codebase

Instead we should figure out a way to use sheets.code4recovery.org to make this possible. We might need to add a special mode to return the refreshed feed contents directly from the request

We should then create a followup issue to create a notice for direct-API users that we are going to deprecate that method, and another followup ticket to remove that logic after, let's say, 6 months

joshreisner:
closes #1348 

enter a Sheet URL (i used [this one](https://docs.google.com/spreadsheets/d/17hg4oj-V_tnRJn5zrRbAAtlEsDM9pN_KeQBUu7QLfy4/edit?gid=0#gid=0) for testing) directly into the importer

leverages new feature in sheets.code4recovery.org - will allow users to do this without signing up for a Google API

show message when using the old, undocumented way (direct API access)

Screenshot 2024-10-26 at 10-16-40 Import   Export ‹ WordPress Test — WordPress

### DIFF
--- a/includes/admin_import.php
+++ b/includes/admin_import.php
@@ -160,23 +160,29 @@ if (!function_exists('tsml_import_page')) {
         if (!empty($_POST['tsml_add_data_source']) && isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce)) {
 
             //sanitize URL, name, parent region id, and Change Detection values
-            $data_source_url = trim(esc_url_raw($_POST['tsml_add_data_source'], array('http', 'https')));
+            $data_source_url = $imported_data_source_url = trim(esc_url_raw($_POST['tsml_add_data_source'], array('http', 'https')));
             $data_source_name = sanitize_text_field($_POST['tsml_add_data_source_name']);
             $data_source_parent_region_id = (int) $_POST['tsml_add_data_source_parent_region_id'];
             $data_source_change_detect = sanitize_text_field($_POST['tsml_add_data_source_change_detect']);
 
+            //use c4r sheets service for Google Sheets URLs
+            if (strpos($data_source_url, 'docs.google.com/spreadsheets') !== false) {
+                $url_parts = explode('/', $data_source_url);
+                $sheet_id = $url_parts[5];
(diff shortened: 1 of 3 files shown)


includes/admin_import.php
```diff
@@ -160,23 +160,29 @@ function tsml_import_page()
         if (!empty($_POST['tsml_add_data_source']) && isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce)) {
 
             //sanitize URL, name, parent region id, and Change Detection values
-            $data_source_url = trim(esc_url_raw($_POST['tsml_add_data_source'], array('http', 'https')));
+            $data_source_url = $imported_data_source_url = trim(esc_url_raw($_POST['tsml_add_data_source'], array('http', 'https')));
```
joshreisner, line 163: `$imported_data_source` is there because i wanted to save the actual sheet URL that the user entered as the feed url, but modify the url that gets imported below

includes/admin_import.php
```diff
@@ -160,23 +160,29 @@ function tsml_import_page()
...
             $data_source_name = sanitize_text_field($_POST['tsml_add_data_source_name']);
             $data_source_parent_region_id = (int) $_POST['tsml_add_data_source_parent_region_id'];
             $data_source_change_detect = sanitize_text_field($_POST['tsml_add_data_source_change_detect']);
 
+            //use c4r sheets service for Google Sheets URLs
+            if (strpos($data_source_url, 'docs.google.com/spreadsheets') !== false) {
+                $url_parts = explode('/', $data_source_url);
+                $sheet_id = $url_parts[5];
+                $imported_data_source_url = 'https://sheets.code4recovery.org/tsml/' . $sheet_id;
+            }
```
gkovats, line 173: Struck me this `sheets.code4recovery.org/tsml/` should be a top level constant at some point.

gkovats (approved):
Nice service!

PR #1559, opened 2024-11-01
protect TSML UI + widgets against display flex

### LINKED ISSUE #1427: Bug in widget layout on some themes with TSML UI appearance
on our website, if i add paragraphs to the top and bottom meetings page widgets using the TSML UI appearance, i see content appearing to the left and right of the meeting finder, when it should be above and below

Screenshot 2024-05-22 at 7 26 05 AM

joshreisner:
close #1427 

GeneratePress (and maybe others) uses `display: flex` on the content parent, which causes TSML UI when used with widgets to show like this:

Screenshot 2024-11-01 at 7 43 36 AM

wrapping it in a div is the simplest fix

Screenshot 2024-11-01 at 7 51 02 AM


gkovats (approved):
`<div>LGTM</div>`

PR #1570, opened 2024-11-16
Add a confirmation before removing a feed

gkovats:
Small, native confirmation before you click on that "X" next to a feed item and find out what happens. Then if you don't have the feed URL on hand, oh well. So, just one step before it's gone.

If we have need for fancier modals, glad to add them, but none stood out.


joshreisner (approved):
good idea

Issue #1569, opened 2024-11-16
Plugin Check "Escape Output" Warnings


joshreisner:
Running WordPress's own Plugin Check plugin shows a lot of warnings.

![Screenshot 2024-11-16 at 08-03-16 Plugin Check ‹ Code for Recovery — WordPress](https://github.com/user-attachments/assets/59564382-ef95-483c-b5dc-f7d47e9e7268)

joshreisner:
going to close this - merged three PRs that handle all of the "Escape Output" warnings. there are many new ones that have popped up in the meantime however. we can document in new issues and handle them in new PRs

PR #1571, opened 2024-11-16
plugin check warnings, part I

joshreisner:
for #1569 

this PR implements a number of recommendations from the plugin check plugin, such as 
* escaping output
* fixing missing translation domains
* adding notes for translators about parameters

additionally it:
* breaks out some functions from functions.php into smaller files (format, get, import)
* lets devsense php plugin reformat whitespace
* removes all one-line if statements
* adds docblocks to all functions
* makes inline commenting style consistent
* removes `tsml_import_reformat_fnv` (obsolete as the FNV database was switched several years ago to FC)
* removes other unused functions

(there are still more warnings, will be handled in subsequent PRs)


includes/functions.php
```diff
@@ -1,64 +1,83 @@
...
+ * add an admin screen update message
+ * used: tsml_import() and admin_types.php
+ * @param mixed $message
+ * @param mixed $type can be success, warning, info, or error
+ * @return void
+ */
 function tsml_alert($message, $type = 'success')
 {
-    echo '<div class="notice notice-' . $type . ' is-dismissible"><p>' . $message . '</p></div>';
+    echo '<div class="notice notice-' . esc_attr($type) . ' is-dismissible"><p>' . esc_html($message) . '</p></div>';
```
gkovats, line 48: ` esc_html($message) ` ends up spiting out `&lt;p&gt;` escaped HTML.

image

includes/functions.php
```diff
@@ -2549,100 +1369,6 @@ function tsml_date_localised($format, $timestamp = null)
...
- * Checks a list of imported meetings for necessary import changes / updates
- *
- * @param array  $feed_meetings                array of feed meetings to test for changes
- * @param string $data_source_url              feed source url
- * @param int    $data_source_parent_region_id feed parent region id
- * @return array[array, array, array]
- *       $import_meetings    array of imported meetings that need to be processes
- *       $delete_meeting_ids array of existing meeting post id's to delete
- *       $change_log         array of changes to existing
- */
```
gkovats, line 2562: Since you're moving this one, maybe move this to `functions_import.php` and rename to `tsml_import_get_changed_meetings()` or something? Since it's very import related.

gkovats (approved):
Chef's kiss. Just on small item about `tsml_alert()`, and one suggestion.

includes/functions.php
joshreisner: ah nice thanks

includes/functions.php
joshreisner: good call

PR #1574, opened 2024-11-22
move early translations to init

### LINKED ISSUE #1573: translation notice in WP 6.7
if WP_DEBUG is not set to `false` you'll see this message

```
Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the 12-step-meeting-list domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.7.0.) in /home/forge/demo.code4recovery.org/public/wp-includes/functions.php on line 6114
```

joshreisner:
fix #1573 

wait for `init` to load translations (`plugins_loaded` happens too soon)

### DIFF
--- a/includes/init.php
+++ b/includes/init.php
@@ -3,6 +3,9 @@
 // for all users
 add_action('init', function () {
 
+    // load text domain
+    tsml_load_config();
+
     // register post types and taxonomies
     tsml_custom_post_types();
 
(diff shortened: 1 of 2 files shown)


gkovats (approved):

gkovats:
Makes sense!

PR #1596, opened 2024-12-21
fix custom types function, meeting detail page display issues

### LINKED ISSUE #1595: custom types not working
our method of adding custom types seems not to be working

```php
if (function_exists('tsml_custom_types')) {
	tsml_custom_types(array(
		'XYZ' => 'My Custom Type',
	));
}
```

best guess is our fix for https://github.com/code4recovery/12-step-meeting-list/issues/1582 broke it

first reported here https://wordpress.org/support/topic/is-there-a-problem-with-custom-type-codes-in-the-current-plugin-release/

also reported in slack

### LINKED ISSUE #1597: [Bug]: Venue Address not appearing on Legacy UI single meeting page
### Contact Details

editor@aatimes.org.au

### Website With Issue

https://meetings.aa.org.au/

### What happened?

Since a recent plugin update, the location address is not appearing in the under the location name on the single meeting page. 
Using Legacy UI
Plugin Version 3.16.8
Example - [https://meetings.aa.org.au/meetings/mill-park-beginners/](https://meetings.aa.org.au/meetings/mill-park-beginners/)
ss01

_No response_

_No response_

joshreisner:
it appears that [this pr](https://github.com/code4recovery/12-step-meeting-list/pull/1574) broke the `tsml_custom_types` function 

context: we were seeing a warning in wordpress that we were calling translations too early - so we moved our program definitions to the `init` hook. however `tsml_custom_types` in a theme's `functions.php` runs before `init`, so those changes were getting overwritten

hopefully this is the right fix for this issue. another option would be to make a global variable for custom types and then import it into our `tsml_load_config()` function and use it at the end to apply the custom types

close #1595

also close #1597 - this was not being output

also fix line breaks not showing up in meeting descriptions

### DIFF
--- a/12-step-meeting-list.php
+++ b/12-step-meeting-list.php
@@ -5,7 +5,7 @@
  * License: GPLv2 or later
  * License URI: https://www.gnu.org/licenses/gpl-2.0.html
  * Description: Manage a list of recovery meetings
- * Version: 3.16.8
+ * Version: 3.16.9
  * Requires PHP: 5.6
  * Author: Code for Recovery
  * Author URI: https://github.com/code4recovery/12-step-meeting-list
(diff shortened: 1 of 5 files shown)


petermixt:
Thank you.  pds

petermixt (approved):
Thank you Josh.
pds

joshreisner:
just released `3.16.9` which should fix this issue

PR #1600, opened 2024-12-25
normalize special characters for change detection

joshreisner:
fix the issue with phantom changes detected in #1599 

need to decode entities when comparing local and remote, or characters like ' show up as `&#39;`


westlinksonline (commented):
Yes, this change makes sense.

ginslo (commented):
Yes, this change makes sense.

PR #1612, opened 2025-01-13
fix distance bug

### LINKED ISSUE #1611: [Bug]: Location filter isn't working
### Contact Details

webmaster@nenyaa.org

### Website With Issue

https://nenyaa.org/meetings/

### What happened?

When I try to narrow my search by location, nothing happens. I put in my location and narrow the range, but get all of the meetings showing no matter how far they are from me. 

IMG_0324
Screenshot_20250104-220958

Chrome

_No response_

joshreisner:
fix #1611 

seems like this value needs to be echoed - i'm guessing the javascript is depending on it to find the currently-selected radius

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -304,6 +304,7 @@ Yes, you will need to know the key name of the field. Then include an array in y
 = 3.16.12 =
 * New meetings block [more info](https://github.com/code4recovery/12-step-meeting-list/pull/1383)
 * Fix custom type descriptions and flags [more info](https://github.com/code4recovery/12-step-meeting-list/issues/1606)
+* Fix "near location" distance bug in legacy UI [more info](https://github.com/code4recovery/12-step-meeting-list/issues/1611)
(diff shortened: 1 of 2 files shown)


gkovats (approved):

Area48Webmaster:
I'm still having the same issue with the location filter. I still won't narrow down the range. It just shows every meeting on a given day.

joshreisner:
huh, i tried searching `Stone Ridge, NY` on your site and i'm seeing it respect the filter now: https://nenyaa.org/meetings/?tsml-day=1&tsml-distance=10&tsml-mode=location&tsml-query=Stone%20Ridge%2C%20NY%2C%20USA

could you tell us what steps  you tried so we can reproduce it the issue?

Area48Webmaster:
Disregard my previous comment. I just checked again and it seems to be working fine. Thank you for all of your help

Issue #1614, opened 2025-01-15
[Bug]: unable to publish Updated greyed out


Oclair:
### Contact Details

feedback@navienna.com

### Website With Issue

https://www.navienna.com

### What happened?

This bug is encountered via the backend.

![image](https://github.com/user-attachments/assets/390d40d2-fae7-4e2a-a381-de20a1fb7188)
While updating an existing meeting, there is no error message printed to indicate why the Update button is deactivated.

Kind Regards,
Jason D

gkovats:
@Oclair  Could you share the details of the Meeting? Usually it'll come down to some logic like Online meetings must have an approximate address. We could definitely improve the messaging.

Oclair:
Definitely an online meeting! [https://www.navienna.com/meetings/dienstag-de/?tsml-day=any](url)

Oclair:
@gkovats hey that's great :)  Thank you for your service!

joshreisner:
yes, thanks @gkovats !

this fix has been released in `3.16.13`

PR #1615, opened 2025-01-15
Fix issue with specific location for online meeting

gkovats:
Addresses issue #1614 

Online meetings with a specific address _shouldn't_ restrict updates to the meeting. Small fix to avoid preventing saves.


joshreisner (approved):
thanks!

PR #1620, opened 2025-01-25
Fix sorting by distance in Legacy UI

### LINKED ISSUE #1619: Distance sorting alphanumerically
see comment here: https://github.com/code4recovery/12-step-meeting-list/discussions/1613#discussioncomment-11936485

not quite sure why this is happening. in public.js it should be `parseFloat`ing the value it's sorting on

https://github.com/code4recovery/12-step-meeting-list/blob/main/assets/src/public.js#L848

joshreisner:
fix #1619 

background: looks like a change to [fix sorting of non-en strings](https://github.com/code4recovery/12-step-meeting-list/issues/146) back in 2020 may not have accounted for the numeric case

i looked into simplifying that code with a `localeCompare` but that doesn't appear to handle polish correctly, so leaving all that as it is


gkovats (approved):
Works on my install

PR #1648, opened 2025-03-26
add homegroup online and finnish

joshreisner:
* Add Finnish [more info](https://github.com/code4recovery/central/issues/31)
* Add Homegroup Online [more info](https://github.com/code4recovery/spec/pull/111)

Screenshot 2025-03-25 at 7 39 57 PM

don't love that the button is too small but wasn't sure if i should take out the `Contribute with` as part of this PR - feel like that can be a followup
Screenshot 2025-03-25 at 7 40 06 PM

Screenshot 2025-03-25 at 7 44 02 PM


gkovats (approved):
Looks good here

tim-rohrer:
Seems to me that with the icon showing bills, the words "Contribute with" could be removed.

joshreisner:
yep that was my first instinct as well @tim-rohrer and called it out in the PR description. however i think that would be best handled in a followup ticket that considered these factors:
* removing "Contribute with" is not great from an accessibility perspective. perhaps the "Contribute with " needs to move to an aria-label for screen readers
* TSML UI is [not as bad with the same change](https://github.com/code4recovery/tsml-ui/pull/436) because it uses different container sizes, perhaps these should be addresses together to have the same approach

these screenshots are the same website / theme / browser / screen size. one potential fix would be to re-evaluate the container size in the legacy detail screen

Screenshot 2025-03-26 at 07-57-42 As Bill Sees It – Test Site

Screenshot 2025-03-26 at 07-58-37 As Bill Sees It

Issue #1580, opened 2024-12-03
Data source name in new column on meetings admin list and in alert on detail page


joshreisner:
### Discussed in https://github.com/code4recovery/12-step-meeting-list/discussions/1578

<div type='discussions-op-text'>

<sup>Originally posted by **jimmedlen** November 29, 2024</sup>
Our Area Website currently has Outside Sources that are imported.
I would like to display the Data Source Name as a column on the WP Meeting and Edit Meeting pages.

https://area22indiana.org/wp-admin/edit.php?post_type=tsml_meeting

https://area22indiana.org/wp-admin/post.php?post=91217&action=edit

The Alert does not identify the Source.
This meeting was imported from an external data source. Any changes you make here will be overwritten when you refresh the data.

![image](https://github.com/user-attachments/assets/7f82f124-a674-40b4-9062-db703cf4c76e)

![image](https://github.com/user-attachments/assets/03c508cf-38fe-4ce9-845c-364d14be7adf)

<img width="435" alt="image" src="https://github.com/user-attachments/assets/a9272218-3322-4df8-9006-cf150455c028">
</div>

joshreisner:
this should be fixed now in `3.16.17`

PR #1661, opened 2025-04-07
Add Data Source labels

gkovats:
Addresses #1580 

- add column for data source to Meetings list
- add label for data source to edit screen
- add a filter for data source to meetings list
image

image

### DIFF
--- a/includes/functions_get.php
+++ b/includes/functions_get.php
@@ -284,7 +284,7 @@ function tsml_get_locations()
  */
 function tsml_get_meeting($meeting_id = false)
 {
-    global $tsml_program, $tsml_programs, $tsml_contact_fields, $tsml_array_fields;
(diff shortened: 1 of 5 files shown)


includes/functions_get.php
```diff
@@ -323,6 +323,8 @@ function tsml_get_meeting($meeting_id = false)
         }
         if (in_array($key, $tsml_array_fields, true)) {
             $value = array_values((array) maybe_unserialize($value));
+        } elseif (in_array($key, $tsml_url_fields, true)) {
```
gkovats, line 326: one note here, the `data_source` url was getting it's entities converted, so `https://site.com/?feed&prop=2` was becoming `https://site.com?feed&amp;prop=2` - ampersand convert. Added logic to avoid this for url fields.

joshreisner (approved):
nice 👏

joshreisner:
also very handy to be able to filter

Issue #1286, opened 2023-12-20
[Feature Request]: Automate Data Source Refreshing


kiyote33:
### Contact Details

waynep@code4recovery.org

### Requested Feature/Enhancement

After the daily Data Source change detection  scan job runs, and it determines that a, database CRUD operation is needed to resync the local database with the records in the incoming remote feed, then the needed operations should automatically take place as part of the scan job. The email sent to the  Change Notification Email recipients should indicate detail information about the records in the database which have been automajically updated.

joshreisner:
this rolled out in `3.17`

PR #1664, opened 2025-04-11
[Issue #1286] Automated imports

gkovats:
Addresses #1286 

Moved a bunch, but really _just_ moved. There were two chunks of logic within `admin_import.php` and `ajax.php` that I had to make standalone functions, so they could be called by the cron event.

- added `tsml_auto_import` setting, which is treated like a boolean (on or off). Perhaps if we want to save more settings later, it'll be an object, but the object is still truthy.
- scheduled event runs every 10 minutes, if a buffer exists pulls 10 items at a time - _should_ be quick, but if it proves slow and drop to 5. If buffer is empty, checks for a data source not updates in last day. Again, another interval worth validating.
- `tsml_import_build_change_report` optionally spits out an array for saving to `tsml_auto_import_log` option, which trims to last 14 days of items
- ajax route to request log in descending date order / reversed
- new admin screen for full log. _Maybe_ not worth a full screen. I dunno. Could be a modal that expands from import / export?

image

image


assets/src/admin.scss
```diff
@@ -2,8 +2,7 @@
 
 body {
 	//import & settings
-	&.tsml_meeting_page_import,
-	&.tsml_meeting_page_settings {
+	.wrap.tsml_admin_settings {
 		@import 'admin-import-settings.scss';
 	}
```
gkovats, line 7: rather than added a third body class, just made a new shared class

includes/ajax.php
```diff
@@ -334,301 +334,28 @@ function tsml_ajax_geocode()
 
 // ajax function to import the meetings in the import buffer
 // used by admin_import.php
-add_action('wp_ajax_tsml_import', function () {
-    global $tsml_data_sources, $tsml_custom_meeting_fields, $tsml_source_fields_map, $tsml_contact_fields, $tsml_entity_fields, $tsml_array_fields;
```
gkovats, line 338: This removed block of logic, responsible for importing meetings from the buffer, is moved to `tsml_import_buffer_next( $count = 25 )`

includes/functions.php
```diff
@@ -340,6 +341,7 @@ function tsml_plugin_deactivation()
     if (post_type_exists('tsml_group')) {
         unregister_post_type('tsml_group');
     }
+    tsml_import_cron_check( false );
```
gkovats, line 344: On plugin deactivate, remove TSML cron jobs. Reminds me to check if we remove cron events for change detection.

includes/admin_import.php
```diff
@@ -156,127 +159,21 @@ function tsml_import_page()
             }
         }
 
-        // add data source
```
gkovats, line 159: This block responsible for polling a new or existing data source for meeting updates and adding them to the buffer now in `tsml_import_data_source($data_source_url, $data_source_name = '', $data_source_parent_region_id = 0, $data_source_change_detect = 'disabled')`

includes/admin_import.php
```diff
@@ -621,104 +527,201 @@ function tsml_import_page()
                     </div>
 
                     <!-- Wheres My Info? -->
```
gkovats, line 529: Just moved export box below where's my info box

includes/functions_import.php
```diff
@@ -1,5 +1,457 @@
 <?php
```
gkovats, line 2: You'll want to check off line, but as I said, the logic is copy pasted from original locations. Only changed the input parameters and the return for the second function.

includes/admin_settings.php
```diff
@@ -15,37 +15,40 @@ function tsml_settings_page()
 
         $tsml_data_sources = tsml_get_option_array('tsml_data_sources');
 
+        // is this a valid TSML post
+        $valid_nonce = isset($_POST['tsml_nonce']) && wp_verify_nonce($_POST['tsml_nonce'], $tsml_nonce);
```
joshreisner, line 19: bizarre in retrospect that we didn't do this sooner

gkovats:
@joshreisner just FYI I expect this PR to stew for a little bit. I still want to do more testing against different import sets, and just make sure there isn't something with importing I upset with the moves.

joshreisner:
broadly: this is really great, thank you for leading this! the importing seems to work great and i fully am on board with the refactoring to the import functions and the way the cron works

the areas that i want to consider are more around the UI:
1. should we include more detail in the activity log to show things like what if a feed is down
2.should we take out change detection and simplify the admin UI
3. should we make automatic imports always-on

joshreisner:
thanks for adding more resolution to the logging - this is looking good. 

upon reflection i think having the setting is how it should be 👍 

let's keep thinking about taking out change detection - perhaps that can be in a subsequent PR but pre-release

joshreisner (approved):
this is great ✅ 

i think we can merge when you're ready and make a few subsequent PRs with refactors before releasing. for example i'd like to explore:

* removing scanning data sources for changes
* visual polish on the logs - maybe color-coding the types
* adding log entries for 
  * editing meetings manually
  * uploading CSV
  * import finish
* showing the import start / finish / error types in the short list (vs meeting changes)

includes/functions_log.php
```diff
@@ -0,0 +1,110 @@
...
+        $tsml_log = tsml_get_option_array('tsml_log');
+        $tsml_log = array_filter($tsml_log, 'is_array');
+        // trim to last 30 days
+        $cutoff = time() - (30 * 24 * 60 * 60);
+        $tsml_log = array_filter($tsml_log, function ($entry) use ($cutoff) {
+            if (!is_array($entry) || !isset($entry['timestamp'])) {
+                return false;
+            }
+            // timestamp can be string (previous) or number (current)
+            $time = is_numeric($entry['timestamp']) ? intval($entry['timestamp']) : strtotime($entry['timestamp']);
```
joshreisner, line 76: cool call switching timestamp to numeric - i think we can simplify some of the logic by throwing out non-numeric entries, since no one is going to miss these invisible geocoding entries

includes/functions_log.php
```diff
@@ -0,0 +1,110 @@
...
+            return $time && $cutoff < $time;
+        });
+        // add updates
+        $tsml_log = array_merge($tsml_log_updates, $tsml_log);
+        update_option('tsml_log', $tsml_log);
+    }
+}
+
+/**
+ * get lot entries
```
joshreisner, line 86: typo "lot"

assets/src/admin.js
```diff
@@ -49,6 +49,77 @@ jQuery(function ($) {
...
+            console.warn(e);
+            $log.find('.log-table__empty').show();
+        })
+        // watch for filter change
+        $filter.on('change',(e) => {
+            const type = $filter.val();
+            $log.find('tr').each((i,e) => $(e).toggle((!type && e.dataset?.type) || type === e.dataset?.type));
+        })
+    }
+
```
joshreisner, line 122: what's the logic in making this a JSON feed? I think it could be less code overall if it were just synchronous php functions, and I see the tsml_log_get() can take arguments

bring this up because we do want to be able to link to a type of log entry so we'd need to add logic here for that - could get unwieldy

includes/functions_log.php
```diff
@@ -0,0 +1,110 @@
...
+        // trim to last 30 days
+        $cutoff = time() - (30 * 24 * 60 * 60);
+        $tsml_log = array_filter($tsml_log, function ($entry) use ($cutoff) {
+            if (!is_array($entry) || !isset($entry['timestamp'])) {
+                return false;
+            }
+            // timestamp can be string (previous) or number (current)
+            $time = is_numeric($entry['timestamp']) ? intval($entry['timestamp']) : strtotime($entry['timestamp']);
+            return $time && $cutoff < $time;
+        });
```
joshreisner, line 78: let's also just limit this to some number (1000?) records

assets/src/admin.js
gkovats: Entirely force of habit, just a pattern I do constantly. And yeah it's extra work, I added some JS var for the edit meeting url format. I'll try moving to part of backend render, should be same result.

includes/functions_log.php
gkovats: Just added `$tsml_log_config['max']` as 1000 default, but allow for future configuration.

assets/src/admin.js
gkovats: Just updated to render entirely on backend, removed the ajax endpoint.

includes/functions_log.php
gkovats: Typos add character. But fixed.

includes/functions_log.php
gkovats: Done, stripping string timestamps.

includes/functions_log.php
joshreisner: ok thanks! seems safer to put it here though instead of in tsml_log_get

joshreisner (approved):

Issue #1669, opened 2025-04-25
[Bug]: Zoom url check needs to account for a new url format


gkovats:
### Contact Details

kovatsgeorge@gmail.com

### What happened?

Informed by #1668, this Zoom url is another valid format that we'll need to support:

`https://us02web.zoom.us/meeting/register/[^\]\/]+#/registration`

Although the check is currently commented out, this should be added to the code in case we re-enable the check.

### Relevant log output/errors

```shell

```

joshreisner:
i believe this shipped in `3.17`

PR #1670, opened 2025-04-26
Accept an alternate Zoom url format

gkovats:
Addresses #1669 , raised by #1668 

Now support 2 types of Zoom urls:
- https://us02web.zoom.us/j/388499177
- https://us02web.zoom.us/meeting/register/tZAucu2pqjIpHtCNCNQ_R3OM7r6m6poOxArr#/registration

Removing attempt to clean up the zoom url, just out of caution - don't know all the supported hash url values and query params. Don't want to deny users their preferred urls.

Because we disabled this check temporarily, I enabled the check when `$tsml_debug` is set to `true` by user via theme.


joshreisner:
the reason to clean up the subdomains is that they are 100% decorative - `us02web.zoom.us` is equivalent to `zoom.us` - but in some cases they can actually imply affiliation or infringe on someone's privacy. for example if i generate a zoom meeting from my work's enterprise account, i will be letting people know where i work

otherwise lgtm, thanks!

gkovats:
Didn't know that, good point! I'll update to strip the subdomain off.

gkovats:
@joshreisner  _finally_ updated, trimming subdomain

joshreisner (approved):

PR #1671, opened 2025-04-26
minor tweaks and refactors for automated importing

joshreisner:
* visual tweaks to log (see below)
* be able to link to a particular filter type
* only show the filter types in use
* hide change detection as an option for new feeds
* only show import log if there are events
* move log page to its own file
* bump version for eventual release
* show the `data_source` type on the import / export page, instead of meeting changes

Screenshot 2025-04-26 at 12 43 04 PM

Screenshot 2025-04-26 at 12 43 39 PM


assets/src/admin.js
```diff
@@ -50,14 +50,18 @@ jQuery(function ($) {
...
             const type = $filter.val();
-            $log.find('tr').each((i,e) => $(e).toggle((!type && e.dataset?.type) || type === e.dataset?.type));
-        })
+            const url = new URL(window.location);
+			if (type) {
+				url.searchParams.set('filter_type', $filter.val());
+			} else if (url.searchParams.has('filter_type')) {
+				url.searchParams.delete('filter_type');
+			}
+			window.location = url.toString();
```
joshreisner, line 63: append a query param (with reload) to be able to link to the page with the type selected

includes/admin_import.php
```diff
@@ -709,81 +685,4 @@ class="button">
         </div>
         <?php
     }
-}
-
-// import log admin page
-function tsml_log_page()
```
joshreisner, line 715: this got moved to a new file

includes/functions_log.php
```diff
@@ -80,8 +80,7 @@ function tsml_log_save_updates()
         $tsml_log = array_merge($tsml_log_updates, $tsml_log);
 
         // upper limit to log entries
-        $max = intval($tsml_log_config['max'] ? $tsml_log_config['max'] : 1000); 
-        $tsml_log = array_slice($tsml_log, 0, $max);
+        $tsml_log = array_slice($tsml_log, 0, 1000);
```
joshreisner, line 83: i simplified this a little. if people need to change this value i think we should just change it globally

assets/src/admin-import-settings.scss
```diff
@@ -34,30 +35,33 @@ ul {
...
+	}
+	&__empty {
+		display: none;
+		td {
+			padding: 1rem !important;
+			text-align: center !important;
+			font-family: inherit !important;
+		}
+	}
 }
```
gkovats, line 65: Something is up with the whitespace config, or lack of config. My editor reads them as spaces. I've used a prettier config to address in other projects, will see if that is appropriate here.

assets/src/admin.js
gkovats: Feel like the ideal set up is filter without reload _and_ persisting selections, but honestly 6 in one hand, half a dozen in the other.

gkovats (approved):
Thanks for cleaning it up Josh. Would honestly be fine with having Devsense extension just auto clean on save. We're coming at this with different code styles from other work, and mine keep creeping in - muscle memory. Appreciate you keep it consistent, otherwise it's harder to work with later.

assets/src/admin.js
joshreisner: yeah - my thinking was to be able to link to the log page with a type seleted - i wasn't sure if i wanted the button on the import / export page to link to just the data source events or to the full log, or to have two buttons

anyway i'm sure we could do both probably as well. the reload didn't seem so bad though since you have to be at the top of the page to use it

PR #1686, opened 2025-06-01
Normalize location info on import

gkovats:
As we do with Group info, this also loops through Locations by Formatted Address and makes sure each Location / `tsml_location` gets the same values. This avoids the issue from Google Sheets where you have the same location with 2 or more values for a field

image

Also moved the timezone sanitization to the same import sanitize function, and tried making it a little more forgiving. I noticed any timezone value in Google Sheets, unless it's exactly an accepted value, causes change notifications.

### DIFF
--- a/includes/functions_import.php
+++ b/includes/functions_import.php
@@ -179,7 +179,7 @@ function tsml_import_data_source($data_source_url, $data_source_name = '', $data
  */
 function tsml_import_buffer_next($limit = 25)
 {
-    global $tsml_data_sources, $tsml_custom_meeting_fields, $tsml_source_fields_map, $tsml_contact_fields, $tsml_entity_fields, $tsml_array_fields;
+    global $tsml_data_sources, $tsml_custom_meeting_fields, $tsml_source_fields_map, $tsml_contact_fields, $tsml_import_fields, $tsml_entity_fields, $tsml_array_fields;
 
     $meetings = tsml_get_option_array('tsml_import_buffer');
     $errors = $remaining = [];
@@ -384,6 +384,7 @@ function tsml_import_buffer_next($limit = 25)
         $custom_meeting_fields = array_merge(
             ['types', 'data_source', 'conference_url', 'conference_url_notes', 'conference_phone', 'conference_phone_notes'],
             array_keys($tsml_source_fields_map),
+            $tsml_import_fields,
             $tsml_entity_fields
         );
         if (!empty($tsml_custom_meeting_fields)) {
@@ -614,6 +615,7 @@ function tsml_import_get_changed_meetings($feed_meetings, $data_source_url)
     // list changed and new meetings found in the data source feed
     foreach ($feed_meetings as $index => $feed_meeting) {
         $feed_meeting_slug = $source_meeting = $source_meeting_id = null;
+        $feed_meeting['import_hash'] = tsml_get_import_hash($feed_meeting);
 
         if (empty($feed_meeting) || !is_array($feed_meeting)) {
             continue;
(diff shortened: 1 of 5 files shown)


includes/functions_import.php
```diff
@@ -1065,21 +1080,49 @@ function tsml_import_sanitize_meetings($meetings, $data_source_url = null, $data
             }
         }
 
-        // track group values
-        if (isset($meetings[$i]['group']) && !empty($meetings[$i]['group'])) {
+        // sanitize timezone if present
+        if (!empty($meetings[$i]['timezone'])) {
+            if (!in_array($meetings[$i]['timezone'], $all_timezones)) {
+                // make an attempt to find timezone
```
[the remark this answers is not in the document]
joshreisner: do we think this is something that could happen with manual entry? feel like if someone were to mess this up they'd be more likely to enter something like "Pacific" or "EST"

includes/functions_import.php
```diff
@@ -1065,21 +1080,49 @@ function tsml_import_sanitize_meetings($meetings, $data_source_url = null, $data
...
+                $timezone = str_replace(' ', '_', $meetings[$i]['timezone']);
+                $matches = array_values(array_filter($all_timezones, function($tz) use ($timezone) {
+                    return false !== stripos($tz, $timezone);
+                }));
+                if (count($matches)) {
```
joshreisner, line 1091: maybe we should say `if (count($matches) === 1) {` to avoid matching the wrong one

joshreisner (approved):
i have a little uncertainty around the partial timezone matching but but overall lgtm, thanks!

probably can keep thinking about timezones - the real question to me is if we want to accept that in the spec and potentially add that to TSML UI

includes/functions_import.php
gkovats: My thinking was if we find more than one match, we pick the first. You think if someone puts something so vague it matches multiple values (count > 1), it should be discarded?

includes/functions_import.php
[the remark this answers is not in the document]
gkovats: Yeah, didn't know how far to go. I'd add some variable list of common aliases if you think it's worth it, and just match for things like `Pacific`, `PST`, `MT`, `MST`, `Mountain`, `Central`... etc.

includes/functions_import.php
joshreisner: yah, it seems dangerous to assume we know if there are several

what if someone put america for example. we would assume america/adak

gkovats: @joshreisner FYI, check the latest. I made a meatier `tsml_timezone_parse()` function that's a little large, but only accepts 1 match case, and then if no matches, will check some canned aliases. Right now it's just US aliases, but could be expanded if we want to torture someone:

https://www.timeanddate.com/time/zones/

joshreisner: thanks! i like that it's more deterministic now

gkovats:
@joshreisner  Sorry, I pushed this last change into this branch, maybe it belongs in it's own PR:
https://github.com/code4recovery/12-step-meeting-list/pull/1686/commits/0a671bdfa68488352c4c87abc6e8de71289b8fa2

It's two things:
- the MD5 hash we discussed today, weird we hadn't done it already!
- added `source_location` = > `location` mapping, another field tracking the incoming "location" name value for granular comparison

Let me know if you want this in a separate PR. I had this branch open, _somewhat_ related

includes/variables.php
```diff
@@ -177,6 +182,7 @@
     'source_region' => 'region',
     'source_sub_region' => 'sub_region',
     'source_slug' => 'slug',
+    'source_location' => 'location',
```
joshreisner, line 185: do we still need this?

gkovats: Nope, removed.

PR #1703, opened 2025-07-02
feat: add support for Marijuana Anonymous

noodnik2:
## Work in Progress

Thanks for the support, Josh!

Feedback from others welcome as well.


joshreisner:
looks good!

PR #1708, opened 2025-07-13
add "clear log" button

joshreisner:
add a button so you can clear the logs 

button

cleared


gkovats (approved):
Works on my machine 👍

PR #1724, opened 2025-08-02
Issue 1710 Alternate Feed Urls

gkovats:
Thinking here is make feeds available via an additional new public url.
`/tsmlfeed/` and `/tsmlfeed/{sharing_key}`

This gets around an odd issue where hosts reject requests to `/wp-admin/admin-ajax.php` under some conditions. It's unclear what was interfering, but it could come up elsewhere. Possible other cases:
- #1684 
- #724

### DIFF
--- a/includes/admin_settings.php
+++ b/includes/admin_settings.php
@@ -478,16 +478,16 @@ if (!function_exists('tsml_settings_page')) {
                                 <?php if (count($tsml_sharing_keys)) { ?>
                                     <table class="tsml_sharing_list">
                                         <?php foreach ($tsml_sharing_keys as $key => $name) {
-                                            $address = admin_url('admin-ajax.php?') . http_build_query([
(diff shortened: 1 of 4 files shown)


joshreisner:
not opposed, but feel like adding rewrite rules could get a little funky  - how about trying something like this? this might be a little more standard since it uses WP's REST API:

```php
<?php

add_action('rest_api_init', function () {
    register_rest_route('tsml/v1', '/meetings', [
        'methods' => 'GET',
        'callback' => 'tsml_rest_endpoint',
        'permission_callback' => 'tsml_rest_permission_check',
    ]);
});

function tsml_rest_endpoint()
{
    $meetings = tsml_get_meetings();
    return rest_ensure_response($meetings);
}

function tsml_rest_permission_check()
{
    global $tsml_sharing;
    return $tsml_sharing === 'open';
}
```

PR #1725, opened 2025-08-03
Add rest endpoint for meeting feed

gkovats:
Adding the rest endpoint for now without announcing in admin. We can discuss next if this replacing the standard path or is shared in the admin as an alternate path.

```
/wp-json/tsml/meetings
/wp-json/tsml/meetings?key={key}
```

### DIFF
--- a/includes/rest.php
+++ b/includes/rest.php
@@ -1,0 +1,27 @@
+<?php
+
+// rest api functions
+
+add_action('rest_api_init', function () {
+    register_rest_route('tsml', '/meetings', [
+        'methods' => 'GET',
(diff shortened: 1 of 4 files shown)


includes/rest.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+// rest api functions
+
+add_action('rest_api_init', function () {
+    register_rest_route('tsml', '/meetings', [
+        'methods' => 'GET',
+        'callback' => 'tsml_rest_feed_endpoint',
```
gkovats, line 8: Left the version out of the path, probably won't have a ton of future endpoints or breaking changes to this one

joshreisner (approved):

PR #1729, opened 2025-08-10
Fix where 11th step type loses key

gkovats:
Addresses #1728 

Maybe I knew this in the past, maybe I'm relearning it now. But `array_merge` doesn't give a damn about your numeric array keys.
https://www.programiz.com/online-compiler/3V497uhDFGKse

So just did tedious loops.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -994,9 +994,11 @@ function tsml_meeting_types($types)
  * return an array of languages, with legacy code substitutions per-program
  * used: variables.php, meeting-edit.php
  */
-function tsml_languages($types = [])
+function tsml_languages($types = [], $program = null)
 {
     global $tsml_program;
+
+    $program = $program ? $tsml_program : $program;
 
     // https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
     $languages = [
@@ -1111,11 +1113,11 @@ function tsml_languages($types = [])
         }
     }
 
-    $languages = array_merge($types, $languages);
-
-    asort($languages);
-
-    return $languages;
+    foreach($languages as $key => $value) {
+        $types[$key] = $value;
+    }
+    asort($types);
+    return $types;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


joshreisner (approved):
thanks for the demo! weird

joshreisner (changes requested):
actually i'm a little worried about reusing $tsml_program

includes/variables.php
```diff
@@ -1432,17 +1432,16 @@ function tsml_load_config()
...
+        $tsml_programs[$tsml_program]['types']['ONL'] = __('Online Meeting', '12-step-meeting-list');
+        $tsml_programs[$tsml_program]['types']['TC'] = __('Location Temporarily Closed', '12-step-meeting-list');
+        $tsml_programs[$tsml_program]['types'] = tsml_languages($tsml_programs[$tsml_program]['types'], );
     }
 
     // remove 'TC' and 'ONL' from default flags if meeting finder is TSML UI
     if ($tsml_user_interface === 'tsml_ui') {
-        foreach ($tsml_programs as $key => $value) {
-            $tsml_programs[$key]['flags'] = array_diff($value['flags'], ['TC', 'ONL']);
+        foreach ($tsml_programs as $tsml_program => $value) {
```
joshreisner, line 1443: i think renaming this probably won't overwrite the `$tsml_program` variable we set on line 607, but only because it's not globalized on line 650 - seems dangerous though

includes/functions.php
```diff
@@ -994,10 +994,12 @@ function tsml_meeting_types($types)
  * return an array of languages, with legacy code substitutions per-program
  * used: variables.php, meeting-edit.php
  */
-function tsml_languages($types = [])
+function tsml_languages($types = [], $program = null)
```
joshreisner, line 997: i also can't quite tell why we're adding the `$program` param here since it doesn't seem like we're using it

joshreisner (approved):
hope you don't mind i went ahead and made those changes. if it looks good feel free to merge or happy to talk more about it

includes/variables.php
gkovats: Oops. Thanks for the catch.

PR #1747, opened 2025-09-10
don't fail if there's a problem with the cache

joshreisner:
for #1746 

improve resiliency if there's a problem json-decoding the meetings cache


SIA-WEBSUP (approved):
assuming that the 'bad cache' is guaranteed to be overwritten down below, this looks good to me.

PR #1752, opened 2025-09-14
sync types with spec

### LINKED ISSUE #924: Sync meeting types with Spec
Spec, TSML, TSML UI, PDF, Sheets, and the Meeting Guide app should all be in sync

Start by doing an analysis of all the products and see where there are discrepancies

Tracking discrepancies in this Google Sheet: https://docs.google.com/spreadsheets/d/1t4uf1poSY_3sspB1I7lwDn6CoTZWdfyDvacfpAk44Oc/edit?usp=sharing

Going to discuss with GSO next week but for now it looks like the actions for TSML are:

* change `Sign Language` to `American Sign Language`
* add `Bisexual`
* change `Online Meeting` to `Online`
* change `Outdoor Meeting` to `Outdoor`
* change `Step Meeting` to `Step Study`

Important to make sure that `Sign Language`, `Outdoor Meeting`, and `Step Meeting` still be supported when importing from CSV

joshreisner:
apologies for the diff - to be more consistent i sorted the AA meeting types by their type code rather than their definition. this is more consistent with the other programs and makes it easier to compare to the types here: https://github.com/code4recovery/spec?tab=readme-ov-file#meeting-types

besides sorting, this PR:
* changes `Outdoor Meeting` to `Outdoor`
* changes `Step Meeting` to `Step Study`
* adds `Bisexual`
* makes it so the old names still work when importing

these apply to AA only

you can test it out locally by using the example spreadsheet, since that still uses the `Step Meeting` type

closes #924


gkovats (approved):
Aliases work as expected from my Googlesheet test, and changes look good!

PR #1782, opened 2025-10-15
fix: redirect legacy ?post_type=tsml_meeting URLs to pretty permalinks

pjaudiomv:
I really don't know this code base well but this is possible solution to https://github.com/code4recovery/12-step-meeting-list/discussions/1781 I tested locally and worked for resolving the error with TSML UI `<Router basename="/meetings/"> is not able to match the URL "/?post_type=tsml_meeting" because it does not start with the basename, so the <Router> won't render anything.`

### DIFF
--- a/includes/init.php
+++ b/includes/init.php
@@ -9,6 +9,25 @@ add_action('init', function () {
     // register post types and taxonomies
     tsml_custom_post_types();
 
+    // Handle legacy redirects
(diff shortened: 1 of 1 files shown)


pjaudiomv:
cool thanks for the additional info, I had not real context for how old those urls were and what not.

PR #1783, opened 2025-10-15
fix rendering on non-default archive pages

joshreisner:
for #1781 

the `data-path` parameter was getting set to `/meetings` since that was the default, but on the legacy `/?post_type=tsml_meeting` it wasn't working

this PR should fix the logic so it only uses `data-path` in the correct situation


pjaudiomv:
Just tested locally, looks good!

PR #1801, opened 2025-10-27
Added docker setup for local dev using wp-env

curtisappl:
image


curtisappl:
I'll open a new PR with these changes

PR #1804, opened 2025-10-28
unset types, fix shortcode

joshreisner:
this PR allows you now to unset meeting types using the `tsml_custom_types()` function per https://github.com/code4recovery/12-step-meeting-list/discussions/1802#discussioncomment-14804651

also this fixes an issue i noticed when working on https://github.com/code4recovery/12-step-meeting-list/discussions/1799#discussioncomment-14794943 - without specifying distance in the shortcode i was seeing the title "Meetings in" - this is because the default was an empty array - appropriate for the pre-hooks version of TSML UI, but now it expects an integer. this wasn't apparent until the [input fix yesterday](https://github.com/code4recovery/tsml-ui/pull/484), because we were ignoring default input.

now `[tsml_ui type="open" mode="me" distance="15"]` yields
Screenshot 2025-10-28 at 6 55 47 AM

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -609,7 +609,7 @@ function tsml_custom_types($types)
     add_action('init', function () use ($types) {
         global $tsml_programs, $tsml_program;
         foreach ($types as $key => $value) {
-            if (empty($value) && array_key_exists($key, $tsml_programs[$tsml_program]['types'])) {
+            if (empty($value)) {
                 // if empty, remove existing type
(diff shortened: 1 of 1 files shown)


pjaudiomv:
> I'm curious if you need to check if the key existing before unsetting? I believe unsetting a non-existing key in an existing array is safe.

ya , using unset() on a non-existent array key is completely safe and won't produce any errors or warnings

PR #1813, opened 2025-11-01
fix/1812-discontinue-appending-unused-types

SIA-WEBSUP:
remove code in tsml_import_sanitize_meetings() that was appending unused meeting type codes to the end of the notes field


joshreisner (commented):
rather not include the copilot-instructions as part of this change

could you also include the version bump prep in this PR? that happens in two places:

1. 12-step-meeting-list.php version bump in two places
2. readme.txt version bump in the header and new row in the changeset with a link to the discussion

thanks!

Issue #1910, opened 2026-04-04
alternative to laravel mix


joshreisner:
### Add a description

we are accumulating some npm vulnerabilities by relying on laravel mix, which is abandoned, and can't be resolved with `npm audit fix`

i'm not certain that Vite is the best option here since this is not a SPA. developers need to be able to run in dev mode when running wordpress locally

i wonder if having no build tool is an option? 

we use mix to create these assets:

1. `admin.min.css` (comples sass, which could be rewritten in CSS)
2. `admin.min.js` (combines and minifies javascript, which we could just not do)
3. `public.min.css` (this is the tricky one, since we import bootstrap and nest it inside the `#tsml {}` SASS seletor)
4. `public.min.js` (combines and minifies javascript, which we could just not do)

```
# npm audit report

elliptic  *
Elliptic Uses a Cryptographic Primitive with a Risky Implementation - https://github.com/advisories/GHSA-848j-6mx2-7j84
No fix available
node_modules/elliptic
  browserify-sign  >=2.4.0
  Depends on vulnerable versions of elliptic
  node_modules/browserify-sign
    crypto-browserify  >=3.4.0
    Depends on vulnerable versions of browserify-sign
    Depends on vulnerable versions of create-ecdh
    node_modules/crypto-browserify
      node-libs-browser  0.4.2 || >=1.0.0
      Depends on vulnerable versions of crypto-browserify
      node_modules/node-libs-browser
        laravel-mix  *
        Depends on vulnerable versions of node-libs-browser
        Depends on vulnerable versions of webpack-dev-server
        node_modules/laravel-mix
  create-ecdh  *
  Depends on vulnerable versions of elliptic
  node_modules/create-ecdh

minimatch  <=3.1.3
Severity: high
minimatch has a ReDoS via repeated wildcards with non-matching literal in pattern - https://github.com/advisories/GHSA-3ppc-4f35-3m26
minimatch has ReDoS: matchOne() combinatorial backtracking via multiple non-adjacent GLOBSTAR segments - https://github.com/advisories/GHSA-7r86-cg39-jmmj
minimatch ReDoS: nested *() extglobs generate catastrophically backtracking regular expressions - https://github.com/advisories/GHSA-23c5-xmqv-rm74
fix available via `npm audit fix --force`
Will install @wordpress/scripts@19.2.4, which is a breaking change
node_modules/markdownlint-cli/node_modules/minimatch
[...]

PR #1911, opened 2026-04-04
Replace Laravel Mix with webpack directly

pjaudiomv:
This is an attempt at #1910 `@wordpress/scripts` already bundles webpack and recommends extending its default config via a `webpack.config.js` file in the project root. The only new dependency is `css-minimizer-webpack-plugin`. This resolves npm vulnerabilities from `laravel-mix` transitive dependencies.

https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/README.md

Alternatively we could go with a build.js script and esbuild, but once i dug into it and started reading the wordpress/scripts docs it made more sense to just continue using webpack without the abstraction of laravel-mix


joshreisner:
thanks for doing this investigation! i agree that we should follow what `@wordpress/scripts` suggests

looks like some of our vulnerabilities are actually coming from them even on the latest version

i'm seeing some javascript errors in legacy frontend - looks like it can't find the functions declared in maps.js - is the script concatenation maybe not working?

also we might want to wait to merge https://github.com/code4recovery/12-step-meeting-list/pull/1868 first

pjaudiomv:
Oh I def had it working but then messed with it and must of broke it, ill check it out and see what issue is. ya some of them are coming from `@wordpress/scripts`

pjaudiomv:
Ok ya.  Laravel Mix's `mix.scripts()`  did plain text concatenation (`mix.js` used webpack). The new webpack config treats each file as a separate module, so functions in `maps.js` aren't visible to `admin.js` and `public.js` Fixed by exposing the four cross-file functions on window at the bottom of maps.js. If we used esbuild/sass directly we could concatenate like laravel mix did.

joshreisner (approved):
this is awesome, thanks

PR #1926, opened 2026-04-18
enable language prop in TSML UI shortcode

joshreisner:
`[tsml_ui]` uses browser language

`[tsml_ui language="es"]` sets the language to Spanish

as discussed in #1846


pjaudiomv (approved):
Nice so this would only set language if explicitly set in short code, else it still defaults back to browser language tsml ui logic

joshreisner:
yah

PR #1969, opened 2026-07-20
applied suggested adm-zip override to package.json to version 0.5.18

ginslo:


joshreisner (approved):
ok i guess we need to remember to remove this once they update their package.json

https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/package.json#L50C1-L51C1

PR #1980, opened 2026-07-29
fix: escape and sanitize event log entries to prevent stored XSS

pjaudiomv:


ginslo:
I see no issues while testing `commit d3052e476c6` on my local test env.

westlinksonline:
Tested locally via WP-CLI. Confirmed input sanitization strips XSS payloads from log entries before storage and admin output is properly escaped. Looks good.

ginslo (approved):
Testing verified (see previous comment from my other account). The sanitization and escaping passed my tests. Approving

pjaudiomv:
thanks @westlinksonline  and @ginslo

PR #1981, opened 2026-07-29
Add `tsml_import_response_error` to improve HTTP response error handling

anchovie91471:
- Introduced `tsml_import_response_error` function to handle and log transport/status errors from data source responses.
- Replaced redundant error-handling code with centralized function calls.
- Enhanced error logging for feed scan failures, ensuring better visibility of issues such as malformed JSON or security challenges.

### DIFF
--- a/includes/functions_import.php
+++ b/includes/functions_import.php
@@ -1,6 +1,47 @@
 <?php
 
 /**
+ * inspect a data source HTTP response for transport and status errors
+ *
+ * Runs before any JSON parsing so that a security challenge or an error page
+ * is reported as such, rather than surfacing as malformed JSON.
+ *
+ * @param mixed $response return value of wp_safe_remote_get()
+ * @return string|null error message, or null when the body is worth parsing
+ */
+function tsml_import_response_error($response)
+{
+    if (is_wp_error($response)) {
+        // translators: %s is the error reported by WordPress
+        return sprintf(__('Could not reach the data source: %s.', '12-step-meeting-list'), $response->get_error_message());
+    }
+
+    if (!is_array($response)) {
+        // translators: %s is the invalid response from the data source
+        return sprintf(__('Invalid response: <pre>%s</pre>.', '12-step-meeting-list'), print_r($response, true));
+    }
+
+    $response_code = (int) wp_remote_retrieve_response_code($response);
+
+    if ($response_code < 200 || $response_code > 299) {
+        // security services answer automated requests with an HTML challenge instead of the feed
+        if (wp_remote_retrieve_header($response, 'cf-ray') || stripos(wp_remote_retrieve_body($response), 'please enable cookies') !== false) {
+            // translators: %s is the HTTP status code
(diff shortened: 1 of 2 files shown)


includes/functions_import.php
```diff
@@ -1,5 +1,46 @@
...
+    if (!is_array($response)) {
+        // translators: %s is the invalid response from the data source
+        return sprintf(__('Invalid response: <pre>%s</pre>.', '12-step-meeting-list'), print_r($response, true));
+    }
+
+    $response_code = (int) wp_remote_retrieve_response_code($response);
+
+    if ($response_code < 200 || $response_code > 299) {
+        // security services answer automated requests with an HTML challenge instead of the feed
+        if (wp_remote_retrieve_header($response, 'cf-ray') || stripos(wp_remote_retrieve_body($response), 'please enable cookies') !== false) {
```
SIA-WEBSUP, line 28: Is the challenge detection  a bit too broad right now?

A Cloudflare cf-ray header can appear on normal 4xx/5xx responses that are not bot/cookie challenges, and the single phrase check seems a little narrow in scope. 

Can we extract this into a dedicated helper that combines multiple signals before classifying as a challenge?

```
<?php
/**
 * Detect likely security challenge/interstitial responses.
 *
 * @param array $response Response array from wp_safe_remote_get().
 * @return bool
 */
function tsml_import_is_security_challenge_response($response)
{
    $status_code = (int) wp_remote_retrieve_response_code($response);
    if ($status_code >= 200 && $status_code <= 299) {
        return false;
    }

    $body = strtolower((string) wp_remote_retrieve_body($response));
    $content_type = strtolower((string) wp_remote_retrieve_header($response, 'content-type'));
    $cf_mitigated = strtolower((string) wp_remote_retrieve_header($response, 'cf-mitigated'));

    // Explicit Cloudflare challenge signal.
    if ($cf_mitigated === 'challenge') {
        return true;
    }

    // Challenges are usually HTML interstitials, not feed payloads.
    $is_html = (strpos($content_type, 'text/html') !== false)
        || (strpos($body, '<!doctype html') !== false)
        || (strpos($body, '<html') !== false);

    if (!$is_html) {
        return false;
    }

    // Keep this list short and high-confidence.
    $markers = [
        '/cdn-cgi/challenge-platform/',
        'cf-browser-verification',
        'checking your browser before accessing',
        'enable javascript and cookies to continue',
        'please enable cookies',
    ];

    foreach ($markers as $marker) {
        if (strpos($body, $marker) !== false) {
            return true;
        }
    }

    return false;
}
```

anchovie91471: Dang! Good catch!

The tsml_import_response_error() already checks for codes 200–299 and it runs before this, so I'd remove the redundant check for those codes:

`if ($status_code >= 200 && $status_code <= 299) {
        return false;
    }`

and change line 28 from

`if (wp_remote_retrieve_header($response, 'cf-ray') || stripos(...'please enable cookies'...) !== false) {`

to

`if (tsml_import_is_security_challenge_response($response)) {`

Sound good?

SIA-WEBSUP: Sounds great!

anchovie91471: Done

SIA-WEBSUP (approved):

PR #1990, opened 2026-08-21
fix: don't let an unreachable data source block the auto-import queue

pjaudiomv:
This bubbled up from https://github.com/code4recovery/12-step-meeting-list/discussions/1989. `tsml_auto_import_check()` refreshes one data source per run, choosing the one with the oldest last_import. last_import is only written on success, so a feed  that cannot be fetched keeps the oldest timestamp forever, gets selected on  every run, and the remaining feeds are never refreshed.

Record last_attempt on both success and failure, and rotate on the later of `last_import` and `last_attempt`. A failing feed is now retried once per 24 hours  like every other feed while last_import continues to report the last successful  sync. The failure is also stored on the data source and shown on the import  screen, since the timestamp alone gives no hint that the feed has been down.

### DIFF
--- a/includes/admin_import.php
+++ b/includes/admin_import.php
@@ -314,7 +314,17 @@ if (!function_exists('tsml_import_page')) {
                                         </td>
 
                                         <td class="align-right">
-                                            <?php echo esc_html(date(get_option('date_format') . ' ' . get_option('time_format'), $properties['last_import'])) ?>
+                                            <?php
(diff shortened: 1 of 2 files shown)


pjaudiomv:
resolves https://github.com/code4recovery/12-step-meeting-list/discussions/1989

joshreisner (approved):

PR #1994, opened 2026-08-27
fix: switch default map tiles from Carto to OpenStreetMap

pjaudiomv:
Carto now watermarks keyless basemap requests with "API KEY REQUIRED". Use OSM standard tiles, which need no key.

see https://github.com/code4recovery/12-step-meeting-list/discussions/1993

https://austinaa.org/meetings/?view=map
https://aasanjose.org/meetings/?view=map

See TSML UI fix here as well https://github.com/code4recovery/tsml-ui/pull/551


tim-rohrer (approved):

joshreisner:
thanks @pjaudiomv and @tim-rohrer

PR #2017, opened 2026-09-17
Refactor internal TSML state access

HamishW5-46:
## Summary

This PR introduces explicit ownership and access for TSML's request-local configuration, WordPress-backed settings, and runtime state. Internal consumers use a small `TSML_State` store and procedural helpers instead of importing loose `$tsml_*` variables with PHP `global` declarations.

Legacy globals remain available for themes, snippets, integrations, and downstream customisations. Translation-sensitive values retain their existing WordPress lifecycle. The change is intended to improve maintainability while preserving the plugin's observable behavior and practical extension interfaces.

## Motivation

TSML historically initialises configuration and settings in `includes/variables.php`, then consumes them throughout the plugin through `global $tsml_*` declarations. This makes dependencies and mutation difficult to follow and mixes declarative configuration with option-backed and translated/runtime values.

It also makes initialization sensitive to PHP include scope. WP-CLI can bootstrap WordPress inside a method, causing bare assignments in an included plugin file to become method-local rather than globally accessible. `05220f1` addresses this with an explicit global declaration in `includes/variables.php`. That working fix is the compatibility baseline for this refactor, not a defect being removed without replacement.

The new store is populated explicitly without relying on included-file assignments becoming PHP globals, and it exports compatibility globals deliberately. This replaces the internal architectural dependence on the large global declaration while retaining the behavior it protects.

## Architecture

`includes/state.php` defines `TSML_State` and the primary procedural access helpers:

```php
$program = tsml_state_get('tsml_program');
$programs = tsml_state_get('tsml_programs');

tsml_state_set('tsml_program', $program, TSML_State::CATEGORY_SETTING);
```

Bulk access uses `tsml_state_get_many()` and `tsml_state_set_many()`. Consumers assign returned keys explicitly; state access does not use `extract()`.

| Category | Ownership and lifecycle | Examples |
| --- | --- | --- |
| Configuration | Definitions/defaults initialized during bootstrap; intentionally mutable where existing customisation APIs require it | Contact/entity fields, URL/array fields, export columns, conference providers, source mappings, timezone aliases, map defaults, Google address overrides |
| WordPress-backed settings | Request-local values initially derived from existing WordPress option APIs; persistence remains explicit through `update_option()` | Selected program, sharing, notification/feedback addresses, cache settings, distance units, automatic imports, timezone, UI selection |
| Runtime state | Values populated or changed during the request and existing hooks | Translated days/strings/programs/types, attendance options, columns, types in use, sort mode, language, nonce, timestamp, reusable handles |
| Compatibility API | Fallback category for registered legacy values imported from external global mutation where no internal category is already assigned |

Categories describe ownership and lifecycle; they do not make configuration immutable. In particular, Google address overrides remain mutable through customisation and geocoding helpers.

Defaults are established in the authoritative initialization path rather than repeatedly supplied by internal consumers. Intentional absence remains supported for optional extension values and the legacy slug override.

The implementation keeps the plugin's procedural WordPress style. It adds no dependency-injection container, autoloading framework, interfaces, namespaces, or broad conversion of plugin functions to classes.

## Lifecycle And Mutation

- The plugin bootstrap loads `includes/state.php` before the existing procedural includes.
- `includes/variables.php` registers compatibility names and explicitly publishes initialized configuration, settings, and runtime values.
- Translation-sensitive configuration remains in `tsml_load_config()` on the existing `init` path. Translated days, strings, programs/types, columns, and attendance options are not promoted to bootstrap-time constants.
- Internal mutation paths explicitly write state, including admin settings, bounds/types updates, customisation helpers, geocoding overrides, and reusable runtime handles.
- WordPress options retain their existing names and persistence APIs. State writes are request-local and do not themselves persist options.
- Existing actions, filters, routes, function entry points, shortcodes, widgets, and block registration are intended to retain their existing behavior and ordering.

## Legacy Global Compatibility

Registered `$tsml_*` globals are retained, including the names protected by the `05220f1` scope fix and practical extension points such as `$tsml_slug`, `$tsml_columns`, `$tsml_sort_by`, `$tsml_map`, `$tsml_defaults`, `$tsml_feedback_url`, `$tsml_google_geocoding_key`, and `$tsml_custom_meeting_fields`.

Synchronization follows explicit rules:

1. A state write immediately exports the value to its compatibility global.
2. A state read checks registered legacy globals against the last exported/imported snapshot.
3. An externally changed registered global is imported when that key is next read through the state API.

Globals are therefore a compatibility surface rather than an independently initialized internal store. Synchronization is access-based, not continuous or equivalent to PHP reference-based `global` semantics in every execution sequence. A local value already copied from state does not observe a later external global mutation until internal code reads the state again.

That limitation needs particular attention in hook-heavy integrations. No blanket claim of reference-level compatibility is made. No legacy globals are proposed for removal or formal deprecation in this PR.

Under `WP_DEBUG`, reading an unregistered state key emits a notice to help catch misspelled keys. This validation reuses state/legacy registration; it does not introduce a separate registry framework or production exception behavior.

## Significant Files

- `12-step-meeting-list.php`: loads the state API before existing includes.
- `includes/state.php`: owns request-local values, categories, bulk access, legacy synchronization, and debug validation.
- `includes/variables.php`: explicitly publishes initialized values and preserves translated configuration loading.
- Admin/import/AJAX/REST/retrieval/save/formatting/email/timezone/shortcode/widget includes: migrate internal reads and mutations to explicit state access.
- `tests/state-loader-scope.php`: adds a standalone regression test for nested-scope configuration loading.
- `.github/workflows/release.yml`: runs the scope regression alongside PHP lint for tagged release builds.
- `.github/workflows/test.yml`: runs PR/push CI coverage.

## Automated Regression Coverage

`php tests/state-loader-scope.php` loads state/configuration from inside a function scope using a small set of WordPress stubs, then verifies:

- A preexisting legacy slug override is retained.
- Contact definitions remain available as an array with the expected 19 fields.
- Compatibility globals are exported after nested-scope loading.
- Sort state initialises to `time`, covering the historical null sort-key failure.
- Runtime configuration loading provides AA programs and attendance options.
- Day ordering honors WordPress's start-of-week setting.
- An external legacy sort-global mutation is imported on the next state read.

This is focused scope/lifecycle coverage, not a full WordPress integration suite. Translation functions are stubbed; real language-plugin behavior and arbitrary hook sequences are not proven by this test.

## Validation Performed

Earlier local validation of the refactor recorded:

- Scope regression: `state-loader-scope-ok`.
- PHP syntax checks across the plugin: passed.
- `git diff --check`: passed at that validation point.
- State/global checks: selected program `aa`, sort `time` in both state and the compatibility global, 19 contact fields, and configuration category for Google overrides.
- Legacy sort-global override and meeting retrieval: successful, with expected number of meetings returned.
- Internal REST request to `/tsml/meetings`: HTTP status 200, array response with expected meeting count and representative expected fields.
- Shortcode smoke checks: meeting/location/region counts, next meetings, types list, regions list, and TSML UI returned output.
- `tsml/meetings` block: registered with attributes and a render callback.
- Origin frontend smoke: meeting UI/configuration/feed markers present in rendered HTML.

**The cron result proves event execution, not a complete import round trip:** the recorded execution was approximately 0.005 seconds. Full new/changed/deleted-meeting import equivalence is unverified.

Checks through browser exercised admin settings, Import & Export, Event Log, regions, districts, meeting lists, and an existing meeting editor. Frontend checks covered combined region/day/attendance/time filters, physical and online details, empty-search recovery, map tiles/markers, calendar menu construction, and a 390 x 844 mobile viewport.

No console warnings or errors were observed on the inspected admin screens. The inspected Apache log had no new entries from the browser checks. `wp-content/debug.log` contained no entries from the inspected checks.

No complete normalised before/after upstream snapshot comparison is claimed. Stable response shape, counts, registration, and representative workflows were inspected, but these checks are not proof of regression-free behavior.

## Review Priorities And Remaining Checks

Please review compatibility synchronisation around hooks that mutate globals after internal functions have already copied state. Also review same-request state writes against upstream mutation semantics; explicit synchronisation should preserve existing global updates and should not silently alter option-only paths. Cache writability is one such path worth checking.

Additional validation remains necessary for full manual/automatic imports, new/changed/deleted meetings and related entities, CRUD persistence, admin settings saves, export contents, alternate sort settings, actual translation/language integrations, and representative filter/action overrides.

Frontend validation should also cover the legacy UI, hybrid meetings, guest sessions, and broader interaction states. Registration/render smoke checks do not establish every frontend interaction.


pjaudiomv:
Hi Hamish, thanks for the interest in 12 Step Meeting List and for the detailed write-up. Speaking just for myself as a contributor, I'm not sure what this buys us in practice. The output is identical, the WP-CLI scope issue was already fixed with one line in `3.19.19`, and the store adds a sync layer on top of the `$tsml_*` globals rather than replacing them. As your write-up notes, that sync is access-based rather than true PHP global semantics, and the globals are a documented extension point a lot of site owners use in their `functions.php`. So it reads to me as more code to maintain without a concrete problem being solved.

If you hit an actual bug or pain point, could you start a thread in the Code for Recovery Slack about it? A smaller, targeted PR would be much easier to review. I'm going to close this for now.

HamishW5-46:
Hi,

Thanks for reviewing it. I understand the concern around adding maintenance surface without a current functional defect beyond the scope issue already addressed in 3.19.19.

Just to clarify one part of my rationale, retaining the globals while decoupling internal consumers was intentional. Given their existing extension points, I didn’t believe removing or deprecating them as part of the same change would be appropriate. The synchronisation layer was intended as a compatibility bridge rather than the eventual architectural goal in itself.

I was mainly interested in exploring whether making internal state ownership explicit would be worthwhile independently of the original WP-CLI bug. 

I’m happy to leave the PR closed. As requested, I'll bring broader ideas like this to Slack first in the future so direction can be discussed, and maybe save my time otherwise.

pjaudiomv:
Thanks Hamish, appreciate you taking it that way, and the rationale makes sense. Definitely wasn't wasted effort from my side, it was a clear write-up and it's useful to have the idea on record if the project ever decides to move that direction.

